Issue #230, opened 2025-06-05
[Bug] 2.1.0 broke MQTT_PROXY_PROTOCOL_ENABLED


Webbeh:
**Describe the bug**
After upgrading to 2.1.0 from 2.0.1 (including 3rd party updates and adding the integration runner), no connection to TBMQ work anymore, with the following exception : 
```
tbmq-1  | 2025-06-05 10:31:29,378 [nioEventLoopGroup-6-10] ERROR o.t.m.b.server.MqttSessionHandler - [a189df60-1d86-436f-8fe9-32198e1c5561] Unexpected Exception
tbmq-1  | io.netty.handler.codec.haproxy.HAProxyProtocolException: header length (over 222) exceeds the allowed maximum (108)
tbmq-1  |       at io.netty.handler.codec.haproxy.HAProxyMessageDecoder.fail(HAProxyMessageDecoder.java:311)
tbmq-1  |       at io.netty.handler.codec.haproxy.HAProxyMessageDecoder.failOverLimit(HAProxyMessageDecoder.java:301)
tbmq-1  |       at io.netty.handler.codec.haproxy.HAProxyMessageDecoder.access$400(HAProxyMessageDecoder.java:33)
tbmq-1  |       at io.netty.handler.codec.haproxy.HAProxyMessageDecoder$HeaderExtractor.extract(HAProxyMessageDecoder.java:389)
tbmq-1  |       at io.netty.handler.codec.haproxy.HAProxyMessageDecoder.decodeLine(HAProxyMessageDecoder.java:292)
tbmq-1  |       at io.netty.handler.codec.haproxy.HAProxyMessageDecoder.decode(HAProxyMessageDecoder.java:246)
tbmq-1  |       at io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:530)
tbmq-1  |       at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:469)
tbmq-1  |       at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:290)
tbmq-1  |       at io.netty.handler.codec.haproxy.HAProxyMessageDecoder.channelRead(HAProxyMessageDecoder.java:228)
tbmq-1  |       at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:444)
[...]

PR #312, opened 2026-05-14
feat(server): honor X-Forwarded-For / X-Real-IP on WS/WSS listeners

dmytro-landiak:
## Pull Request description

Adds an opt-in mechanism for the WS (8084) and WSS (8085) MQTT listeners to honor HTTP forward headers (`X-Forwarded-For`, `X-Real-IP`) on the WebSocket upgrade request, so the real client IP is reflected in session info, logs, and the unauthorized-clients feature when TBMQ sits behind a reverse proxy that cannot deliver the PROXY protocol on its HTTP leg (e.g. NGINX `http`/`location` blocks).

Closes the remaining WS/WSS gap discussed in #230.

**Configuration (opt-in, defaults to `false`):**
- `listener.ws.forward_headers_enabled` — env `MQTT_WS_FORWARD_HEADERS_ENABLED`
- `listener.wss.forward_headers_enabled` — env `MQTT_WSS_FORWARD_HEADERS_ENABLED`

**Behavior:**
- A new `ForwardHeadersIpAddressHandler` is inserted into the WS/WSS Netty pipeline immediately after `HttpObjectAggregator`. On the WebSocket upgrade request it reads `X-Forwarded-For` (leftmost trimmed token) → falls back to `X-Real-IP`, parses the value with Guava `InetAddresses.forString` (supports bare and bracketed IPv6), and overrides `MqttSessionHandler.ADDRESS` with `new InetSocketAddress(parsedIp, 0)` (port `0` — the real client port is not carried by either header). The handler then removes itself from the pipeline.
- Falls back silently to the raw socket address if no usable header is present (empty, `unknown`, malformed, etc.) — a misconfigured proxy must not break connectivity.
- PROXY protocol always takes precedence: when proxy_protocol is enabled for the listener (per-listener or via the global `listener.proxy_enabled`), the new handler is not installed.
- HAProxy and other PROXY-protocol setups are untouched: defaults are `false`, and the precedence rule guarantees identical behavior even if both flags are mistakenly enabled.

**Security:**
Documented in the YAML comments. Enabling `forward_headers_enabled` while TBMQ is reachable directly (no trusted proxy in front) allows clients to spoof their source IP via the upgrade headers — operators must only enable it behind a reverse proxy that overwrites these headers before forwarding the request.

**Tests:**
- `ForwardHeadersIpAddressHandlerTest` — 11 unit tests via `EmbeddedChannel` covering: single XFF IP, XFF chain (leftmost wins), only X-Real-IP, both headers (XFF wins), empty / `unknown` / malformed values, IPv6 (bare and bracketed), no headers, and the non-`FullHttpRequest` defensive path.
- `AbstractMqttWsChannelInitializerWiringTest` — 3 pipeline-wiring tests verifying the handler is present when forward_headers is on and proxy is off, and absent when proxy is on or forward_headers is off.

**Documentation:** A short note will need to be added to the TBMQ documentation describing the new env vars and the security caveat. Out of scope for this PR (lives in the website / docs repo).

## General checklist

## Back-End feature checklist


Webbeh:
Thank you!

Issue #259, opened 2025-09-23
[Question] Support for Subscribing to Device Connection/Disconnection Events via Integration Executor?


yiqiqq2025:
Description:
Summary
Is it possible to use the Integration Executor in TBMQ to subscribe to and forward device connection/disconnection (online/offline) status change events? For example, capturing events from internal Kafka topics like tbmq.client.session.event.request (e.g., CONNECTION_REQUEST or DISCONNECTION_REQUEST) and routing them to external systems such as HTTP endpoints, Kafka, or another MQTT broker?
Background

TBMQ Version: [2.1.0]
Environment: [Docker, single-node]
Current Setup: We have HTTP integrations working for regular MQTT PUBLISH messages (e.g., via topic filters like tbmq/#), but connection/disconnection events are not triggering forwards since they are control packets, not publishes.
Workaround Attempted: Using LWT for offline events works for publishes, but we'd prefer a broker-level subscription for all connect/disconnect events (including graceful and unexpected disconnections) without modifying device code.

Desired Functionality

Subscribe to internal session events via Integration Executor.
Filter and transform events (e.g., extract client_id, event_type, timestamp).
Forward to external targets (e.g., POST to HTTP with JSON payload: {"deviceId": "device123", "status": "online", "timestamp": 1695335123456}).
Any configuration examples or limitations (e.g., ACL requirements, supported targets)?

Questions

Is this supported out-of-the-box in current versions (e.g., via custom integration configs or YAML rules)?
If not, are there plans to add it, or workarounds like custom plugins/extensions?
Relevant docs or code examples? (E.g., extending the executor to poll Kafka topics.)

Additional Context

This would enable real-time device fleet monitoring without relying on client-side LWT for all scenarios.
References: [Integration Executor docs](https://thingsboard.io/docs/mqtt-broker/integrations/), [Architecture overview](https://thingsboard.io/docs/mqtt-broker/architecture/).

Issue #127, opened 2024-06-19
How to help business systems perceive whether credentials are online in real time


baigod:
**Component**

* Core

**Description**
We maintain a list of devices in our business system and create unique MQTT credentials for each device, matching them one-to-one with TBMQ credentials. Now, we want our business system to be able to know in real-time whether a device is online. What are the best practices or solutions for interacting with TBMQ to achieve this?

**Environment**

* OS:  docker
* TBMQ: 1.3.0

dmytro-landiak:
Hi @baigod!

Thank you for reaching out! You can achieve real-time monitoring of device statuses using the TBMQ API. The API allows you to check the online status of a device by querying its client session. Here's an example of how to do this using a curl command:

```
curl -X 'GET' \
  'https://HOSTNAME/api/client-session?clientId=CLIENT_ID' \
  -H 'accept: application/json' \
  -H 'X-Authorization: Bearer TOKEN'

```

Replace the HOSTNAME to your actual one and the CLIENT_ID parameter with the actual client ID of the device you want to check. The X-Authorization header should contain a valid JWT token for authentication.

In addition, you can choose another API like below:

```
curl -X 'GET' \
  'https://HOSTNAME/api/client-session?pageSize=100&page=0' \
  -H 'accept: application/json' \
  -H 'X-Authorization: Bearer TOKEN'
```

If you need further assistance or have any questions, feel free to ask!

baigod:
Thank you for your reply, but I think I prefer events like Connect Event and Disconnect Event in thingsboard. I can proactively inform the business system through Kafka or HTTP. Based on my estimation, when I have millions of devices connected to TBMQ, if I call through HTTP https://HOSTNAME/api/client-session The interface and timeliness may suffer significant losses.
I have an idea. Can I subscribe to topics such as tbmq. client. session through a new subscription group to consume tbmq's session events, so that the business system can perceive real-time changes in device connection status.

dmytro-landiak:
@baigod,

Thank you for your feedback. Your approach of subscribing to `tbmq.client.session` topic for real-time session events is indeed better suited for your use case, especially with a large number of devices.

You can connect to Kafka with a separate process, creating a new consumer group to read from the tbmq.client.session topic. Please be aware of the payload format. We use Protobuf for serialization. You can find the object definition for ClientSessionInfoProto [here](https://github.com/thingsboard/tbmq/blob/main/common/queue/src/main/proto/queue.proto#L95).

[...]

PR #323, opened 2026-06-11
feat(integration): deliver client lifecycle events to integrations

deaflynx:
## Pull Request description

Adds the ability for integrations to receive **client lifecycle events**, in addition to the regular published messages they already process. Until now an integration could only be fed MQTT `PUBLISH` messages matched by topic filters; this PR lets an integration also be notified when a client connects, disconnects, (un)subscribes, or fails authentication/authorization, and deliver those events to the external system over the existing HTTP / Kafka / MQTT integration types.

Requested by the community in #259 and #127.

### Event types

| Event type | When it is emitted | Type-specific payload fields |
|---|---|---|
| `CLIENT_CONNECTED` | Client connection accepted | `cleanStart`, `keepAlive`, `protocolVersion`, `sessionExpiryInterval` |
| `CLIENT_DISCONNECTED` | Client session ends (incl. cross-node takeover) | `disconnectReason` (MQTT reason-code name, e.g. `SESSION_TAKEN_OVER`) |
| `CLIENT_SUBSCRIBED` | Subscriptions granted | `subscriptions[]` — each with `topicFilter`, `qos`, `options` (`noLocal`/`retainAsPublish`/`retainHandling`), plus `shareName` and `subscriptionId` when present |
| `CLIENT_UNSUBSCRIBED` | Subscriptions actually removed | `subscriptions[]` — each with `topicFilter`, plus `shareName` when the removed subscription was shared |
| `CLIENT_AUTHENTICATION_FAILED` | Authentication fails | `reason` |
| `CLIENT_AUTHORIZATION_FAILED` | A publish/subscribe is denied | `action` (`publish`/`subscribe`), `topic` |
| `CLIENT_CONNECTION_FAILED` | Broker refuses the connection (e.g. quota exceeded, server unavailable) — not auth failures | `reason` |

Every event also carries the common fields `eventType`, `clientId`, `username`, `clientCertCn`, `sessionId`, `ipAddress`, `ts`, `tbmqNode`, and the integration `metadata`. String fields are omitted when empty (e.g. `username` when authentication is disabled, `clientCertCn` when the client did not authenticate with an X.509 certificate).

### Configuration / delivery modes

An integration's configuration now has two independent, optional opt-in lists:

- `topicFilters` — data messages to deliver (existing behavior).
- `lifecycleEventTypes` — lifecycle event types to deliver (new).

An integration may specify either or both, which yields three delivery modes: **data-only** (unchanged), **events-only**, and **data + events**. At least one of the two lists must be non-empty.

### Design highlights

- **Dedicated per-integration events stream.** Lifecycle events flow over their own Kafka topic, separate from the data topic, with its own (shorter) retention since events are best-effort hints. The data topic remains byte-identical `PublishIntegrationMsgProto`, so the existing data path is untouched.
- **Per-node opt-in cache.** A reverse index (`event type → interested integration ids`) is built at startup and kept in sync via internode notifications on integration save/delete, so the connect/disconnect/sub/unsub hot path does an O(1), allocation-free lookup and produces nothing when no integration is interested.
- **Best-effort production.** The lifecycle-event send runs on the MQTT processing thread (the same synchronous path a regular publish already uses) and is exception-safe: failures are swallowed, counted via a dropped-events metric, and never affect MQTT processing. A bounded `max.block.ms` caps the worst-case stall.
- **Generified IE pipeline.** The integration-executor consume/ack machinery was made generic over the message type so a single implementation serves both the data and the events streams; each integration runs a dedicated events consumer only when it opts into lifecycle events.

### Backward compatibility

Fully backward compatible — no upgrade script required:

- All new proto fields are additive; lifecycle events use a separate stream, so a rolling upgrade has no poison-pill hazard on the data topic.
- Existing data-only integrations behave exactly as before.

### Documentation

- Integration documentation should describe the new `lifecycleEventTypes` option, the available event types, and the JSON payload for each.
- The new dedicated lifecycle-events Kafka topic and its producer/consumer settings (see `thingsboard-mqtt-broker.yml`, `queue.kafka.integration-msg.*event*`) should be documented.

## General checklist

## Back-End feature checklist


dmytro-landiak (commented):
## Review summary

Reviewed 98 changed files in [Feat/mqtt events integration](https://github.com/thingsboard/tbmq/pull/323). This adds a dedicated, best-effort per-integration lifecycle-events stream (connect / disconnect / subscribe / unsubscribe / auth-failure / authz-deny) with a per-node event-type cache kept fresh via internode notifications, opt-in delivery modes, and the supporting UI.

The implementation is defensively coded and I found no correctness bugs: every emission is wrapped in try/catch with a dropped-event metric, returns early when no integration is opted in, and the disconnect/authorization paths guard a null `sessionInfo`. Left 5 comments inline, all minor — a test-coverage gap on the subscribe path, a DRY refactor for the duplicated config parsing, and a few small cleanliness/maintainability notes.

---

application/src/test/java/org/thingsboard/mqtt/broker/actors/client/service/handlers/MqttSubscribeHandlerTest.java
```diff
@@ -99,6 +100,8 @@ public class MqttSubscribeHandlerTest {
     ApplicationPersistenceProcessor applicationPersistenceProcessor;
     @MockitoBean
     RateLimitService rateLimitService;
+    @MockitoBean
+    IntegrationLifecycleEventPublisher integrationLifecycleEventPublisher;
```
dmytro-landiak: This wires in the publisher mock, but I don't see assertions for the two new subscribe-side emissions: `publishSubscribed` should fire (with only the granted subscriptions) on a successful SUBACK, and `publishAuthorizationDenied(ctx, "subscribe", topic)` should fire once per denied topic. The parallel unsubscribe path got a dedicated test (`givenSubscribedAndNeverSubscribedFilters_whenProcess_thenEmitOnlyRemovedFilters`) — worth adding the symmetric subscribe coverage so a future refactor of `collectMqttReasonCodes` / `subscribeAndPersist` can't silently drop these events.

application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/BrokerInitializer.java
```diff
@@ -109,6 +116,28 @@ public void onApplicationEvent(ApplicationReadyEvent event) {
...
 
+    void initIntegrationLifecycleEventCache() {
+        List<Integration> integrations = integrationService.findAllIntegrations();
+        for (Integration integration : integrations) {
+            JsonNode configuration = integration.getConfiguration();
+            if (configuration == null || !configuration.has("lifecycleEventTypes")) {
+                continue;
+            }
+            Set<ClientLifecycleEventType> eventTypes = new HashSet<>();
+            configuration.get("lifecycleEventTypes").forEach(node -> {
```
dmytro-landiak: The "parse the `lifecycleEventTypes` JSON array into a `Set<ClientLifecycleEventType>`, warning on unknown names" logic is now repeated here, in `InternodeNotificationsConsumerImpl.applyIntegrationLifecycleConfig`, and in `InternodeNotificationsServiceImpl.applyIntegrationLifecycleConfig`; and the `"lifecycleEventTypes"` key is hardcoded here, in `DefaultTbIntegrationService.toLifecycleConfigProto`, `DefaultPlatformIntegrationService.updateSubscriptions`, and the DAO validator (the IE side already has `IntegrationEventsOptInUtil.LIFECYCLE_EVENT_TYPES`). Could the key constant and the array→enum-set parse be pulled into one shared helper? It would keep the unknown-value handling consistent and remove several copies to maintain.

application/src/main/java/org/thingsboard/mqtt/broker/service/integration/DefaultPlatformIntegrationService.java
```diff
@@ -122,15 +122,28 @@ public void processServiceInfo(ServiceInfo serviceInfo) {
     @Override
     public void updateSubscriptions(Integration integration) {
         JsonNode configuration = integration.getConfiguration();
-        if (!configuration.has("topicFilters")) {
-            log.error("[{}][{}] Topic filters not configured", integration.getId(), integration.getName());
+
+        boolean hasTopicFilters = configuration.has("topicFilters") &&
```
dmytro-landiak: Minor: `configuration.has("topicFilters")` and `configuration.get("topicFilters")` traverse the node twice (same for `lifecycleEventTypes` just below). Grabbing the node once reads a little cleaner — `JsonNode tf = configuration.get("topicFilters"); boolean hasTopicFilters = tf != null && tf.isArray() && !tf.isEmpty();` — and it matches how the DAO validator already does it.

application/src/main/java/org/thingsboard/mqtt/broker/service/integration/IntegrationLifecycleEventPublisherImpl.java
```diff
@@ -0,0 +1,224 @@
...
+                    .setTopic(topic)
+                    .build();
+            publish(integrationIds, proto);
+        } catch (Throwable t) {
+            onPublishError(ClientLifecycleEventType.CLIENT_AUTHORIZATION_FAILED, t);
+        }
+    }
+
+    private ClientLifecycleEventMsgProto.Builder newSessionBuilder(ClientSessionCtx ctx, ClientLifecycleEventType eventType) {
+        SessionInfo sessionInfo = ctx.getSessionInfo();
```
dmytro-landiak: `newSessionBuilder` dereferences `sessionInfo` and `sessionInfo.getClientInfo()` without a null check. It's safe today — every caller either runs after the session is established or guards `getSessionInfo() == null` first, and the outer try/catch would absorb an NPE into a dropped-event count anyway — but the precondition is implicit. A one-line javadoc noting "callers must ensure sessionInfo is set" would keep a future caller from quietly turning every event into a dropped-and-logged NPE.

application/src/main/java/org/thingsboard/mqtt/broker/server/MqttSessionHandler.java
```diff
@@ -109,10 +109,13 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) {
             address = getAddress(ctx);
             clientSessionCtx.setAddress(address);
         }
+        if (clientSessionCtx.getChannel() == null) {
```
dmytro-landiak, line 112: Moving from setting the channel on every message to capturing it once when null is a nice optimization, given the `ChannelHandlerContext` is stable for the life of a per-connection handler. Just confirming there's no path where the same `ClientSessionCtx` is reused with a different `ctx` (e.g. a pipeline rebuild) where the stale channel would now be retained — I don't think there is, but it's a behavioral change to a core handler worth a second look.

application/src/main/resources/thingsboard-mqtt-broker.yml
```diff
@@ -620,6 +620,15 @@ queue:
...
       additional-producer-config: "${TB_KAFKA_IE_MSG_ADDITIONAL_PRODUCER_CONFIG:}"
+      # Kafka topic properties separated by semicolon for the dedicated `tbmq.ie.event` lifecycle-event topics.
+      # Retention defaults shorter than the data topic since events are best-effort hints.
+      event-topic-properties: "${TB_KAFKA_IE_EVENT_MSG_TOPIC_PROPERTIES:retention.ms:86400000;segment.bytes:26214400;retention.bytes:104857600;partitions:1;replication.factor:1}"
+      # Additional Kafka consumer configs separated by semicolon for the `tbmq.ie.event` topics.
+      additional-event-consumer-config: "${TB_KAFKA_IE_EVENT_MSG_ADDITIONAL_CONSUMER_CONFIG:max.poll.records:50}"
+      # Additional Kafka producer configs separated by semicolon for the `tbmq.ie.event` topics. Configured non-blocking
+      # (low max.block.ms) so a slow or unavailable Kafka can never stall an MQTT connect/disconnect/subscribe/unsubscribe
+      # thread; events are dropped and counted instead.
+      additional-event-producer-config: "${TB_KAFKA_IE_EVENT_MSG_ADDITIONAL_PRODUCER_CONFIG:max.block.ms:5000}"
```
dmytro-landiak: This send isn't actually non-blocking from the caller's perspective. `IntegrationEventMsgQueuePublisherImpl.sendEventMsg` -> `TbPublishServiceImpl.send` -> `TbKafkaProducerTemplate.send` calls `producer.send(record, callback)` directly on the calling thread, and `KafkaProducer.send()` can block up to `max.block.ms` (5s here) while fetching topic metadata or when the buffer is full. Those callers run on MQTT hot-path threads — `publishDisconnected` on the client actor thread, `publishConnected` on the connect executor, `publishAuthorizationDenied` on the publish/subscribe path — so a slow or unavailable Kafka can stall MQTT processing for up to 5s, not "never" as the comment above states. Worth either softening the comment to the real bound or offloading the send to a dedicated executor so the hot path truly never waits on Kafka.

application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/handlers/MqttUnsubscribeHandler.java
```diff
@@ -43,20 +45,46 @@ public class MqttUnsubscribeHandler {
...
+    private List<String> getRemovedTopicFilters(String clientId, List<String> requestedTopics) {
+        Set<TopicSubscription> currentSubscriptions = clientSubscriptionService.getClientSubscriptions(clientId);
+        if (CollectionUtils.isEmpty(currentSubscriptions)) {
+            return List.of();
+        }
+        Set<String> currentTopicFilters = currentSubscriptions.stream()
+                .map(TopicSubscription::getTopicFilter)
+                .collect(Collectors.toSet());
+        return requestedTopics.stream()
+                .filter(topic -> currentTopicFilters.contains(toTopicFilter(topic)))
```
dmytro-landiak: For shared subscriptions this returns the raw requested filter (e.g. `$share/group/topic`), since `requestedTopics` keeps the original strings and only the membership check is normalized through `toTopicFilter`. But `publishSubscribed` emits `TopicSubscription.getTopicFilter()`, which is the bare filter (`topic`) with the share name stripped. So the same shared subscription shows up as `topic` in CLIENT_SUBSCRIBED but `$share/group/topic` in CLIENT_UNSUBSCRIBED — not quite the symmetry the method comment promises. Returning `toTopicFilter(topic)` instead of the raw `topic` would keep the two events consistent.

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/validation/PublishMsgValidationServiceImpl.java
```diff
@@ -51,6 +53,7 @@ boolean validateClientAccess(ClientSessionCtx ctx, String clientId, String topic
         if (!isClientAuthorized) {
             log.warn("[{}][{}][{}] Client is not authorized to publish to the topic {}",
                     clientId, ctx.getSessionId(), ctx.getAuthRulePatterns(), topic);
+            integrationLifecycleEventPublisher.publishAuthorizationDenied(ctx, "publish", topic);
```
dmytro-landiak: This fires a lifecycle event on every denied publish, and (per the publisher note) the Kafka send is synchronous on this validation thread. A client that keeps publishing to a topic it isn't authorized for generates one event — and one synchronous Kafka send — per attempt, both amplifying onto the events topic and repeatedly paying the publish-path latency. It's bounded by the client's publish rate limit and best-effort by design, so this may well be acceptable — just flagging that per-denial emission with no dedup is the behavior a looping/misbehaving client will hit. Was that the intended trade-off?

common/integration/integration-api/src/main/java/org/thingsboard/mqtt/broker/integration/api/AbstractIntegration.java
```diff
@@ -181,6 +185,74 @@ protected ObjectNode constructBody(PublishIntegrationMsgProto msg) {
...
+        body.put("eventType", msg.getEventType());
+        body.put("clientId", msg.getClientId());
+        body.put("sessionId", msg.getSessionId());
+        body.put("ipAddress", msg.getIpAddress());
+        body.put("ts", msg.getTs());
+        body.put("tbmqNode", msg.getTbmqNode());
+        body.put("username", msg.getUsername());
+
+        switch (msg.getEventType()) {
+            case "CLIENT_CONNECTED":
```
dmytro-landiak: This switch keys off raw string literals (`"CLIENT_CONNECTED"`, ...) even though the canonical `ClientLifecycleEventType` enum already lives in `common/data` and this module already depends on it. The proto carries the type as a String, so the literals here and the enum names can drift independently — rename or add a value on one side and this builder silently falls through to `default`. Switching on `ClientLifecycleEventType.valueOf(msg.getEventType())` (or comparing against `.name()`) would keep the two coupled; the publisher side already builds the proto from the enum's `.name()`.

application/src/main/java/org/thingsboard/mqtt/broker/service/integration/IntegrationLifecycleEventPublisher.java
```diff
@@ -0,0 +1,38 @@
...
+
+    void publishDisconnected(ClientSessionCtx ctx, DisconnectReasonType reasonType);
+
+    void publishSubscribed(ClientSessionCtx ctx, List<TopicSubscription> subscriptions);
+
+    void publishUnsubscribed(ClientSessionCtx ctx, List<String> topicFilters);
+
+    void publishAuthenticationFailed(ClientSessionCtx ctx, String clientId, String reason);
+
+    void publishAuthorizationDenied(ClientSessionCtx ctx, String action, String topic);
```
dmytro-landiak: `action` is a free-form String, and the call sites pass the magic literals `"publish"` / `"subscribe"` (in `PublishMsgValidationServiceImpl` and `MqttSubscribeHandler`); it then flows untouched into the proto and the emitted JSON body. A two-value enum (e.g. `PUBLISH`/`SUBSCRIBE`) would make the contract self-documenting and impossible to misspell, versus relying on every caller to type the right lowercase string.

dmytro-landiak (commented):
## Review summary

Reviewed 99 changed files in [Feat/mqtt events integration](https://github.com/thingsboard/tbmq/pull/323). This adds a dedicated, per-integration, best-effort lifecycle-events stream (connect/disconnect/sub/unsub/auth) alongside the existing data stream, with a per-node opt-in cache kept in sync via internode notifications, an executor-side events consumer, and UI controls. The design is clean and well-tested — generifying the IE pack-processing pipeline to share one consume/ack/commit loop across both streams, the lock-free reverse-index cache, and the broad unit coverage all stood out. Left 12 comments inline. The one worth a closer look is that the lifecycle send is synchronous on MQTT hot-path threads and can block up to `max.block.ms`, despite the config comment claiming it can never stall; the rest are mostly maintainability/cleanliness and a couple of test-coverage gaps.

---

application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/connect/ConnectServiceImpl.java
```diff
@@ -295,6 +300,9 @@ boolean shouldProceedWithConnection(ClientActorStateInfo actorState, MqttConnect
             validateLastWillMessage(ctx, clientId, msg);
         } catch (ConnectionValidationException e) {
             log.warn("[{}] Connection validation failed: {}", ctx.getSessionId(), e.getMessage());
+            // ctx.getSessionInfo() is not set yet at this point, so pass the already-built sessionInfo explicitly.
+            // reason is the MQTT connect return code the client received, e.g. CONNECTION_REFUSED_CLIENT_IDENTIFIER_NOT_VALID.
+            integrationLifecycleEventPublisher.publishConnectionFailed(ctx, sessionInfo, e.getMqttConnectReturnCode().name());
```
dmytro-landiak, line 307: The `reason` on `CLIENT_CONNECTION_FAILED` ends up carrying two different vocabularies depending on the path: this pre-connection validation path passes `MqttConnectReturnCode` names (e.g. `CONNECTION_REFUSED_CLIENT_IDENTIFIER_NOT_VALID`), while `refuseConnection` passes internal `ClientSessionFailureReason` names (`SERVER_UNAVAILABLE` / `QUOTA_EXCEEDED` / `UNSPECIFIED_ERROR`). A consumer parsing `reason` then has to understand both schemes for one event type. It's a little surprising next to the sibling change in this same diff that deliberately normalized `disconnectReason` to MQTT-standard names via `MqttReasonCodeResolver` for exactly that portability reason. If unifying both paths onto one vocabulary is more than you want here, documenting the two forms on the proto's `reason` comment would at least keep external consumers from being caught out.

integration/executor/src/main/java/org/thingsboard/mqtt/broker/integration/service/processing/backpressure/IntegrationEventAckStrategyConfiguration.java
```diff
@@ -0,0 +1,36 @@
...
+ * Ack strategy for the dedicated lifecycle-event stream. Kept as a distinct type (rather than a second
+ * binding of {@link IntegrationAckStrategyConfiguration}) so both configs inject unambiguously, and so events
+ * can stay best-effort ({@code SKIP_ALL}) independently of the data stream's strategy.
+ */
+@Component
+@ConfigurationProperties(prefix = "queue.integration-msg.event-ack-strategy")
+@Data
+public class IntegrationEventAckStrategyConfiguration {
+
+    private IntegrationAckStrategyType type;
```
dmytro-landiak: This class is field-for-field identical to `IntegrationAckStrategyConfiguration`. The javadoc reasonably justifies a distinct type (unambiguous injection + independent defaults), so this isn't wrong — but did you consider a shared base holding `type`/`retries`/`pauseBetweenRetries`, with the two subclasses supplying only the `@ConfigurationProperties` prefix? A future field would then be added once instead of in two places. Minor — just flagging the copy so the two don't drift.

dmytro-landiak (commented):
## Re-review summary

Re-reviewed [feat(integration): deliver client lifecycle events to integrations](https://github.com/thingsboard/tbmq/pull/323) — verified 5 finding(s) from the [previous review](https://github.com/thingsboard/tbmq/pull/323#pullrequestreview-4600886417). The fix diff since then spans 16 commits.

| Status | Count |
|--------|-------|
| ✅ Resolved | 4 |
| 💬 Acknowledged | 1 |

All previous findings have been addressed. Also left 2 new low-severity design notes on the fix commits, inline.

### Previous-finding details

- ✅ **application/src/main/resources/thingsboard-mqtt-broker.yml** — the events producer comment claimed the send could "never" stall an MQTT thread — *Resolved: the comment now documents the real up-to-`max.block.ms` stall bound.*
- ✅ **application/.../actors/client/service/handlers/MqttUnsubscribeHandler.java** — shared-sub UNSUBSCRIBE emitted `$share/group/topic` while CLIENT_SUBSCRIBED emitted the bare filter — *Resolved: now emits the bare topic filter plus `shareName`, symmetric with CLIENT_SUBSCRIBED.*
- 💬 **application/.../service/mqtt/validation/PublishMsgValidationServiceImpl.java** — a lifecycle event fires on every denied publish with no dedup — *Acknowledged: confirmed intentional (best-effort, bounded by the client's publish rate limit) and now documented in code.*
- ✅ **common/integration/.../integration/api/AbstractIntegration.java** — the body builder switched on raw string literals that could drift from the enum — *Resolved: now switches on the `ClientLifecycleEventType` enum via a lenient `fromName`.*
- ✅ **application/.../service/integration/IntegrationLifecycleEventPublisher.java** — `action` was a free-form String the callers hand-typed — *Resolved: introduced the `AuthorizationAction` enum (`publish`/`subscribe`).*

---

PR #335, opened 2026-07-03
fix(stats): deregister appProcessor per-client meters on clear to stop registry leak

dmytro-landiak:
## Pull Request description

**Problem.** The per-Application-client processing counters (`appProcessor.*` — 8 counters tagged with `clientId`, plus the same set under the shared-subscription compound id `clientId_shareName_topicFilter`) are registered in the Micrometer registry but were never removed when a client disconnected. `clearApplicationProcessorStats` / `clearSharedApplicationProcessorStats` only dropped the local `managedApplicationProcessorStats` map entry, and `DefaultCounter.clear()` only zeroes the local `AtomicInteger`. As a result, every Application client / shared subscription that ever connected left its counters permanently in the registry, so both the Micrometer registry (heap) and the `/actuator/prometheus` scrape grew without bound under client churn.

**Fix.** Added `StatsFactory.remove(DefaultCounter)` (implemented in `DefaultStatsFactory` via `MeterRegistry.remove(...)`), and the clear paths in `StatsManagerImpl` now deregister each of the cleared client's 8 counters. The 3 `appProcessor.latency` timers are intentionally left registered: they carry no `clientId` tag, so Micrometer keeps a single shared set aggregated across all Application clients — removing them on one client's disconnect would deregister timers still used by the others.

**Scope of changes:**
- `common/stats`: new `StatsFactory.remove(DefaultCounter)` + `DefaultStatsFactory` implementation; `DefaultCounter.getMicrometerCounter()` accessor.
- `application`: `StatsManagerImpl` clear paths deregister the per-client counters; class javadoc on `DefaultApplicationProcessorStats` documenting that the latency timers are intentionally global (no `clientId` tag — do not add one, cardinality trap); `thingsboard-mqtt-broker.yml` comment on `stats.application-processor.enabled` clarifying the per-client cardinality and the global latency aggregation.

The `APPLICATION_PROCESSOR_STATS_ENABLED` default is unchanged (`true`).

**Documentation notes.** No user-facing configuration keys changed. The `stats.application-processor.enabled` yml comment was expanded. If broker metrics are documented externally, it is worth reflecting that `appProcessor.latency` is a global aggregate (not per-client) and that the per-client `appProcessor.*` counters are deregistered on client disconnect.

## General checklist

_Not applicable — back-end only, no UI/API changes._

## Back-End feature checklist

- [x] If new dependency was added: the dependency tree is checked for conflicts. _(No new dependency.)_

New tests (`StatsManagerImplTest`): per-client counters removed from the registry on clear; shared-subscription compound counters removed on shared clear; the global `appProcessor.latency` timers are retained after clear.

### DIFF
--- a/application/src/test/java/org/thingsboard/mqtt/broker/service/stats/StatsManagerImplTest.java
+++ b/application/src/test/java/org/thingsboard/mqtt/broker/service/stats/StatsManagerImplTest.java
@@ -1,0 +1,86 @@
+/**
+ * Copyright © 2016-2026 The Thingsboard Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.thingsboard.mqtt.broker.service.stats;
+
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+import org.junit.Before;
+import org.junit.Test;
+import org.springframework.test.util.ReflectionTestUtils;
+import org.thingsboard.mqtt.broker.common.stats.DefaultStatsFactory;
+import org.thingsboard.mqtt.broker.common.stats.StatsFactory;
+import org.thingsboard.mqtt.broker.common.stats.StatsType;
+import org.thingsboard.mqtt.broker.service.subscription.shared.TopicSharedSubscription;
+
+import static org.junit.Assert.assertEquals;
+import static org.thingsboard.mqtt.broker.common.stats.StatsConstantNames.CLIENT_ID_TAG;
+
+public class StatsManagerImplTest {
+
+    private static final String APP_PROCESSOR = StatsType.APP_PROCESSOR.getPrintName();
+    private static final String APP_PROCESSOR_LATENCY = APP_PROCESSOR + ".latency";
+
+    private SimpleMeterRegistry meterRegistry;
+    private StatsManagerImpl statsManager;
+
+    @Before
+    public void setUp() {
+        meterRegistry = new SimpleMeterRegistry();
+        StatsFactory statsFactory = new DefaultStatsFactory(meterRegistry);
+        statsManager = new StatsManagerImpl(statsFactory);
+        ReflectionTestUtils.setField(statsManager, "applicationProcessorStatsEnabled", true);
+    }
+
+    private int appProcessorCounters(String clientId) {
+        return meterRegistry.find(APP_PROCESSOR).tag(CLIENT_ID_TAG, clientId).counters().size();
+    }
+
+    @Test
+    public void givenApplicationProcessorStats_whenClear_thenPerClientCountersRemovedFromRegistry() {
+        String clientId = "app-client-1";
+        statsManager.createApplicationProcessorStats(clientId);
+        assertEquals(8, appProcessorCounters(clientId));
+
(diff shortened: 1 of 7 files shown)


common/stats/src/main/java/org/thingsboard/mqtt/broker/common/stats/DefaultCounter.java
```diff
@@ -28,6 +28,10 @@ public DefaultCounter(AtomicInteger aiCounter, Counter micrometerCounter) {
         this.micrometerCounter = micrometerCounter;
     }
 
+    public Counter getMicrometerCounter() {
```
dmytro-landiak: This accessor's only caller is `DefaultStatsFactory.remove`, which lives in the same package (`org.thingsboard.mqtt.broker.common.stats`). Making it `public` widens exposure of the raw Micrometer `Counter` to the whole codebase for no current benefit — package-private visibility would keep encapsulation tight while still letting `DefaultStatsFactory` reach it. The `remove(DefaultCounter)` parameter type itself is fine, for what it's worth — it mirrors the factory's existing concrete return types like `createDefaultCounter`/`createStatsCounter`, so accepting the concrete class is consistent. It's just this getter's reach that could be narrowed.

application/src/test/java/org/thingsboard/mqtt/broker/service/stats/StatsManagerImplTest.java
```diff
@@ -0,0 +1,86 @@
...
+        statsManager.createApplicationProcessorStats(clientId);
+        assertEquals(8, appProcessorCounters(clientId));
+
+        statsManager.clearApplicationProcessorStats(clientId);
+
+        assertEquals(0, appProcessorCounters(clientId));
+    }
+
+    @Test
+    public void givenSharedApplicationProcessorStats_whenClearShared_thenCompoundCountersRemovedFromRegistry() {
```
dmytro-landiak, line 72: The three clear paths in `StatsManagerImpl` aren't all covered here: `clearApplicationProcessorStats(clientId)` and this `clearSharedApplicationProcessorStats(clientId)` are tested, but the third overload `clearSharedApplicationProcessorStats(clientId, TopicSharedSubscription)` isn't. It has its own distinct branching — it removes a single compound id from the per-client set and only drops the outer set entry once that set becomes empty. Could you add a case that creates two shared subscriptions for one client, clears one by `(clientId, subscription)`, and asserts that only that compound id's 8 counters are deregistered while the other's remain (and that the outer set entry survives until the last one is cleared)? That's the branch most likely to break silently in a future refactor.

dmytro-landiak (commented):
## Review summary

Reviewed 7 changed files in [fix(stats): deregister appProcessor per-client meters on clear to stop registry leak](https://github.com/thingsboard/tbmq/pull/335). The fix is correct and well-scoped: `StatsCounter extends DefaultCounter`, so `getStatsCounters().forEach(statsFactory::remove)` deregisters exactly the 8 per-client counters; removal happens after the final "Stats on clear" log line so nothing is lost; the disabled path returns the stub and never populates the map (clear is a safe no-op); and all three clear call sites are teardown/cancellation paths, so deregistering matches "entity is gone" semantics. The latency-timer retention is correct and well-documented. Left 2 comments inline (a visibility nit and a test-coverage gap), plus one additional finding about a sibling leak the new `remove` API could also address.

## Additional findings

- **common/integration/cluster-integration-api/src/main/java/org/thingsboard/mqtt/broker/service/DefaultIntegrationStatisticsService.java:114** — The per-integration stats path looks like it has the same registry leak this PR fixes. `IntegrationProcessorStatsImpl` registers 8 `StatsCounter`s tagged with `integrationId`, but `clearIntegrationProcessorStats` / `clearIntegrationEventProcessorStats` route through `disable(...)`, which only flips `active = false`; `printProcessorStats` later drops the entry from the managed map, yet nothing ever calls `meterRegistry.remove`, so the per-integration Micrometer series stay registered across integration create/delete churn. The new `StatsFactory.remove` javadoc even names "an integration" as a use case, so leaving this path untouched reads as a conspicuous skip. Integration churn is admin-driven and far lower-rate than client churn, so it's lower-severity — but it'd be worth either applying the new `remove` API here too or noting in the PR why it's deferred.

---

PR #336, opened 2026-07-03
fix(stats): deregister integration processor meters on clear to stop registry leak

dmytro-landiak:
## Pull Request description

**Problem.** The per-integration processing counters — `integrationProcessor.*` and `integrationEventProcessor.*` (8 `StatsCounter`s each, tagged with `integrationId`) — are registered in the Micrometer registry but never removed. `clearIntegrationProcessorStats` / `clearIntegrationEventProcessorStats` only flip `active = false` (`disable(...)`); the next `printStats()` logs a final line and drops the entry from the managed map, but `meterRegistry.remove(...)` is never called. So on integration create/delete churn (each recreate mints a fresh UUID) the per-integration Micrometer series accumulate for the process lifetime — a registry/cardinality leak on the integration-executor, doubled by the event-processor stream (16 series per integration).

This is the same class of leak fixed for the application-processor counters in #335, and reuses the `StatsFactory.remove(DefaultCounter)` primitive added there.

**Fix.** `printProcessorStats` now deregisters the counters at the point the inactive entry is atomically dropped from the managed map. Removal is tied to the map removal outcome (`computeIfPresent(...) == null`), so if the same integration id is re-enabled during the ≤ one-print-interval window between clear and the next print, its freshly-registered meters are kept rather than stranded:

```java
if (!stats.isActive()) {
    boolean removed = managed.computeIfPresent(stats.getIntegrationUuid(),
            (id, oldStats) -> oldStats.isActive() ? oldStats : null) == null;
    if (removed) {
        stats.getStatsCounters().forEach(statsFactory::remove);
    }
}
```

Applies to both the message stream (`integrationProcessor.*`) and the event stream (`integrationEventProcessor.*`).

**Scope of changes:**
- `common/integration/cluster-integration-api`: `DefaultIntegrationStatisticsService.printProcessorStats` deregisters per-integration counters on the inactive-clear drop.
- New unit tests (`DefaultIntegrationStatisticsServiceTest`).

**A related item that turned out NOT to need a change.** A metrics audit flagged a possible broker-side `stats.ie.enabled` gating gap (the `integration_stats_counter`/`_gauge` lifecycle metrics not materializing on a broker node). On investigation this is correct by design: those metrics are emitted only via the integration lifecycle state machine (`handleIntegrationLifecycleMsg` → `processStart`/`processStop` → `onIntegrationStateUpdate`/`onIntegrationsCountUpdate`), whose only non-test caller runs on the integration-executor — exactly where `stats.ie.enabled` is defined. The broker only performs integration validation/CRUD and never emits these metrics. No gating/config change is made.

**Documentation notes.** No user-facing configuration keys changed; no documentation update required.

## General checklist

_Not applicable — back-end only, no UI/API changes._

## Back-End feature checklist

- [x] If new dependency was added: the dependency tree is checked for conflicts. _(No new dependency.)_

New tests (`DefaultIntegrationStatisticsServiceTest`): per-integration counters removed from the registry after clear + print for both the message and event streams; active (uncleared) integration counters retained after print; clearing one integration leaves another's counters intact.

### DIFF
--- a/common/integration/cluster-integration-api/src/main/java/org/thingsboard/mqtt/broker/service/DefaultIntegrationStatisticsService.java
+++ b/common/integration/cluster-integration-api/src/main/java/org/thingsboard/mqtt/broker/service/DefaultIntegrationStatisticsService.java
@@ -201,20 +201,29 @@ public class DefaultIntegrationStatisticsService implements IntegrationStatistic
                     .map(statsCounter -> statsCounter.getName() + " = [" + statsCounter.get() + "]")
                     .collect(Collectors.joining(" "));
             log.info("[{}][{}] {}: {}", printName, stats.getIntegrationUuid(), label, msgStatsStr);
-            if (!stats.isActive()) {
-                log.trace("[{}] Clearing inactive Integration stats", stats.getIntegrationUuid());
-                boolean removed = managed.computeIfPresent(stats.getIntegrationUuid(),
-                        (id, oldStats) -> oldStats.isActive() ? oldStats : null) == null;
-                if (removed) {
-                    // Deregister the per-integration counters so they stop being scraped and don't leak the
-                    // Micrometer registry on integration create/delete churn. Tied to the atomic map removal
-                    // above so a same-id re-enable in the meantime keeps its freshly-registered meters.
-                    stats.getStatsCounters().forEach(statsFactory::remove);
-                }
-            } else {
+            if (stats.isActive()) {
                 stats.reset();
+            } else {
+                removeInactiveStats(managed, stats);
             }
         }
+    }
+
+    private void removeInactiveStats(Map<UUID, IntegrationProcessorStats> managed, IntegrationProcessorStats stats) {
+        log.trace("[{}] Clearing inactive Integration stats", stats.getIntegrationUuid());
+        // Drop the entry and deregister its per-integration counters together, inside the remap so both
+        // happen atomically under the map's per-key lock. This stops the counters being scraped and bounds
+        // the Micrometer registry on integration create/delete churn, while keeping the meters of a same-id
+        // integration that was re-enabled in the meantime (seen active here) instead of stranding them.
+        // Removal is deferred to this print cycle (not eager on clear, as StatsManagerImpl does for the
+        // application-processor stats) so an inactive integration still emits one final stats line above.
+        managed.computeIfPresent(stats.getIntegrationUuid(), (id, oldStats) -> {
+            if (oldStats.isActive()) {
+                return oldStats;
+            }
+            oldStats.getStatsCounters().forEach(statsFactory::remove);
+            return null;
+        });
     }
 
     @Override
--- a/common/integration/cluster-integration-api/src/main/java/org/thingsboard/mqtt/broker/service/DefaultIntegrationStatisticsService.java
+++ b/common/integration/cluster-integration-api/src/main/java/org/thingsboard/mqtt/broker/service/DefaultIntegrationStatisticsService.java
@@ -211,19 +211,18 @@ public class DefaultIntegrationStatisticsService implements IntegrationStatistic
 
     private void removeInactiveStats(Map<UUID, IntegrationProcessorStats> managed, IntegrationProcessorStats stats) {
         log.trace("[{}] Clearing inactive Integration stats", stats.getIntegrationUuid());
-        // Drop the entry and deregister its per-integration counters together, inside the remap so both
-        // happen atomically under the map's per-key lock. This stops the counters being scraped and bounds
-        // the Micrometer registry on integration create/delete churn, while keeping the meters of a same-id
-        // integration that was re-enabled in the meantime (seen active here) instead of stranding them.
-        // Removal is deferred to this print cycle (not eager on clear, as StatsManagerImpl does for the
-        // application-processor stats) so an inactive integration still emits one final stats line above.
-        managed.computeIfPresent(stats.getIntegrationUuid(), (id, oldStats) -> {
-            if (oldStats.isActive()) {
-                return oldStats;
-            }
-            oldStats.getStatsCounters().forEach(statsFactory::remove);
-            return null;
-        });
+        // Drop the entry only while it is still inactive: a same-id re-enable that already replaced it with an
+        // active entry is kept (the remap returns that value, so removed == false), so its meters aren't touched.
+        // Keep the remap trivial and deregister the counters afterwards, outside the per-key lock: MeterRegistry
+        // #remove takes the registry lock and fires removed-listeners, which is more than a ConcurrentHashMap
+        // remap should do while holding the bin lock. This mirrors how StatsManagerImpl deregisters its
+        // per-client counters. Removal is deferred to this print cycle (not eager on clear) so an inactive
+        // integration still emits one final stats line above.
+        boolean removed = managed.computeIfPresent(stats.getIntegrationUuid(),
+                (id, oldStats) -> oldStats.isActive() ? oldStats : null) == null;
+        if (removed) {
+            stats.getStatsCounters().forEach(statsFactory::remove);
+        }
     }
 
     @Override
--- a/common/integration/cluster-integration-api/src/main/java/org/thingsboard/mqtt/broker/service/DefaultIntegrationStatisticsService.java
+++ b/common/integration/cluster-integration-api/src/main/java/org/thingsboard/mqtt/broker/service/DefaultIntegrationStatisticsService.java
@@ -203,7 +203,14 @@ public class DefaultIntegrationStatisticsService implements IntegrationStatistic
             log.info("[{}][{}] {}: {}", printName, stats.getIntegrationUuid(), label, msgStatsStr);
             if (!stats.isActive()) {
                 log.trace("[{}] Clearing inactive Integration stats", stats.getIntegrationUuid());
-                managed.computeIfPresent(stats.getIntegrationUuid(), (clientId, oldStats) -> oldStats.isActive() ? oldStats : null);
+                boolean removed = managed.computeIfPresent(stats.getIntegrationUuid(),
+                        (id, oldStats) -> oldStats.isActive() ? oldStats : null) == null;
+                if (removed) {
+                    // Deregister the per-integration counters so they stop being scraped and don't leak the
+                    // Micrometer registry on integration create/delete churn. Tied to the atomic map removal
+                    // above so a same-id re-enable in the meantime keeps its freshly-registered meters.
+                    stats.getStatsCounters().forEach(statsFactory::remove);
+                }
             } else {
                 stats.reset();
             }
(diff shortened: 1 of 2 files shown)


dmytro-landiak (commented):
## Review summary

Reviewed 2 changed files in [fix(stats): deregister integration processor meters on clear to stop registry leak](https://github.com/thingsboard/tbmq/pull/336). This is the integration-side counterpart to #335, and the core fix is correct: the `ie-stats-service` scheduler is single-threaded so `printStats` can't race itself, `clear*` only flips `active=false`, and `computeIfPresent(...) == null` correctly gates the deregistration to the entry it actually dropped. `IntegrationProcessorStatsImpl` holds only the 8 per-integration counters (no shared timers), so `getStatsCounters().forEach(statsFactory::remove)` fully cleans up. Left 3 comments inline — a low-probability race plus an untested branch around the re-enable guard, structural/DRY alignment with the sibling `StatsManagerImpl` fix, and a test-parameterization suggestion.

---

common/integration/cluster-integration-api/src/main/java/org/thingsboard/mqtt/broker/service/DefaultIntegrationStatisticsService.java
```diff
@@ -203,7 +203,14 @@ private void printProcessorStats(Map<UUID, IntegrationProcessorStats> managed, S
             log.info("[{}][{}] {}: {}", printName, stats.getIntegrationUuid(), label, msgStatsStr);
             if (!stats.isActive()) {
                 log.trace("[{}] Clearing inactive Integration stats", stats.getIntegrationUuid());
-                managed.computeIfPresent(stats.getIntegrationUuid(), (clientId, oldStats) -> oldStats.isActive() ? oldStats : null);
+                boolean removed = managed.computeIfPresent(stats.getIntegrationUuid(),
```
dmytro-landiak: The reason removal is tied to `computeIfPresent(...) == null` rather than an unconditional remove+deregister is to preserve the meters of a same-id integration that was re-enabled in the meantime — but that guard is worth a second look on two fronts.

First, it's untested: the four tests only hit create→clear→print (`removed`) and create→print (active/reset). None reaches the case where the loop sees an inactive snapshot entry while the map already holds a re-registered *active* entry for that id (`removed==false`), which is the entire justification for the guard. It could be silently regressed (e.g. simplified to an unconditional remove) with every test still green.

Second, the guard only closes the window where the re-enable happens *before* this `computeIfPresent`. `create*`/`clear*` run on integration-executor threads (`IntegrationMsgProcessorImpl`), concurrently with this scheduler thread — so a `createIntegrationProcessorStats(sameId)` landing *between* this decision and the `forEach(statsFactory::remove)` two lines below would re-obtain the same meter ids and then have them deregistered out from under it, stranding an active integration's counters. The window is tiny (two adjacent statements, no yield) and integration churn is admin-driven, so this is very low probability — but given the branch is both untested and not fully closed, is the extra complexity earning its keep versus a simpler approach, or worth hardening plus a targeted test (e.g. reflectively seeding the map state)?

common/integration/cluster-integration-api/src/main/java/org/thingsboard/mqtt/broker/service/DefaultIntegrationStatisticsService.java
```diff
@@ -203,7 +203,14 @@ private void printProcessorStats(Map<UUID, IntegrationProcessorStats> managed, S
...
+                        (id, oldStats) -> oldStats.isActive() ? oldStats : null) == null;
+                if (removed) {
+                    // Deregister the per-integration counters so they stop being scraped and don't leak the
+                    // Micrometer registry on integration create/delete churn. Tied to the atomic map removal
+                    // above so a same-id re-enable in the meantime keeps its freshly-registered meters.
+                    stats.getStatsCounters().forEach(statsFactory::remove);
```
dmytro-landiak: This is now the second place solving the same "per-entity counters leak the registry on churn" problem, and the two diverge in structure: `StatsManagerImpl` removes from the managed map eagerly in `clear*` and prints+deregisters synchronously in `printAndRemoveApplicationStatsOnClear`, whereas here `clear*` only flips `active=false` and the actual removal+deregister is deferred to the next print cycle. The deregister line and its ~3-line comment are near-duplicated across both files.

The loop body here now does several things — format+log the line, then either `reset()` (active) or drop the map entry *and* deregister the meters (inactive) — a fair amount behind a name that only promises `print*`. Would extracting the inactive-cleanup branch into a small private helper (mirroring the sibling's `printAndRemoveApplicationStatsOnClear`) read better and make the two leak-fixes structurally parallel? And if the deferred-print-vs-eager-on-clear divergence is deliberate (e.g. so an inactive integration emits one final stats line before being dropped), a one-line note would help the next maintainer avoid "aligning" one to the other by mistake.

common/integration/cluster-integration-api/src/test/java/org/thingsboard/mqtt/broker/service/DefaultIntegrationStatisticsServiceTest.java
```diff
@@ -0,0 +1,103 @@
...
+        assertThat(counters(INTEGRATION_PROCESSOR, integrationId)).isEqualTo(8);
+
+        service.clearIntegrationProcessorStats(integrationId);
+        service.printStats();
+
+        assertThat(counters(INTEGRATION_PROCESSOR, integrationId)).isZero();
+    }
+
+    @Test
+    void givenIntegrationEventProcessorStats_whenClearedAndPrinted_thenPerIntegrationCountersRemovedFromRegistry() {
```
dmytro-landiak: These four test methods are near-identical create/clear/print/assert scaffolding differing only by `StatsType` and the create/clear pair, and the coverage is asymmetric: the message stream gets `active retained` and `two integrations, one cleared`, but the event stream only gets the clear→removed case. Since both streams run through the exact same `printProcessorStats`, a single `@ParameterizedTest` over the two `StatsType`s (with the corresponding create/clear method references) would both collapse the duplication and give the event stream the same active-retained / two-integration parity in one move.

common/integration/cluster-integration-api/src/main/java/org/thingsboard/mqtt/broker/service/DefaultIntegrationStatisticsService.java
```diff
@@ -201,15 +201,31 @@ private void printProcessorStats(Map<UUID, IntegrationProcessorStats> managed, S
...
+        // happen atomically under the map's per-key lock. This stops the counters being scraped and bounds
+        // the Micrometer registry on integration create/delete churn, while keeping the meters of a same-id
+        // integration that was re-enabled in the meantime (seen active here) instead of stranding them.
+        // Removal is deferred to this print cycle (not eager on clear, as StatsManagerImpl does for the
+        // application-processor stats) so an inactive integration still emits one final stats line above.
+        managed.computeIfPresent(stats.getIntegrationUuid(), (id, oldStats) -> {
+            if (oldStats.isActive()) {
+                return oldStats;
+            }
+            oldStats.getStatsCounters().forEach(statsFactory::remove);
```
dmytro-landiak: The remapping function passed to `computeIfPresent` now performs an external side effect (deregistering meters) rather than just deriving a new value. That's deliberate here — it's what makes drop-entry + deregister atomic and closes the re-enable race, so I think it's the right call. Just flagging the trade-off: `ConcurrentHashMap`'s contract asks remap functions to be short since they run under the per-bin lock, and `statsFactory.remove` walks the registry to unregister. Might be worth a one-line note that the side effect is intentional and must stay cheap, so a future change doesn't grow the work done here under the lock.

common/integration/cluster-integration-api/src/test/java/org/thingsboard/mqtt/broker/service/DefaultIntegrationStatisticsServiceTest.java
```diff
@@ -0,0 +1,169 @@
...
+        IntegrationProcessorStats registered = service.createIntegrationProcessorStats(integrationId);
+        int registeredCounters = registered.getStatsCounters().size();
+        assertThat(counters(INTEGRATION_PROCESSOR, integrationId)).isEqualTo(registeredCounters);
+
+        // Simulate a concurrent re-enable landing between printStats' values() snapshot and the atomic
+        // remap: the entry reports inactive on the snapshot-loop check (so cleanup runs), but by the time
+        // the remap re-reads it the entry is active again, so its meters must be kept, not deregistered.
+        IntegrationProcessorStats reEnabling = mock(IntegrationProcessorStats.class);
+        when(reEnabling.getIntegrationUuid()).thenReturn(integrationId);
+        when(reEnabling.isActive()).thenReturn(false, true);
```
dmytro-landiak, line 162: This stub encodes that `printProcessorStats` calls `isActive()` exactly twice, in order (snapshot-loop check, then the remap re-read). It's a clever way to deterministically hit a branch that's otherwise only reachable under real concurrency, and I don't see a cleaner option. The durability risk: if `printProcessorStats` is later refactored to read `isActive()` a different number of times (caches it, adds a guard), this test would silently take the wrong path or throw without obviously pointing back to the cause. A short comment on the stub — "first call = snapshot-loop check, second = remap re-read; keep in sync with printProcessorStats" — would help the next maintainer, since the coupling can't really be avoided.

dmytro-landiak (commented):
## Re-review summary

Re-reviewed [fix(stats): deregister integration processor meters on clear to stop registry leak](https://github.com/thingsboard/tbmq/pull/336) — verified 3 finding(s) from the [previous review](https://github.com/thingsboard/tbmq/pull/336#pullrequestreview-4624987397), all addressed in `98f9c95a4`.

| Status | Count |
|--------|-------|
| ✅ Resolved | 3 |

Also found 2 minor, non-blocking maintainability suggestions on the fix commit, commented inline.

### Finding details

- ✅ **DefaultIntegrationStatisticsService.java** — re-enable guard: narrow race + untested branch — *Deregistration moved inside the `computeIfPresent` remap, so drop-entry and deregister are now atomic under the per-key lock (closes the race); the new `givenSameIdReEnabledBetweenSnapshotAndRemap` test locks the keep branch.*
- ✅ **DefaultIntegrationStatisticsService.java** — second copy of the leak-fix / loop doing too much — *Extracted `removeInactiveStats`, with the atomic-removal and deferred-vs-eager (vs `StatsManagerImpl`) rationale documented.*
- ✅ **DefaultIntegrationStatisticsServiceTest.java** — test duplication + event-stream coverage asymmetry — *Parameterized over both streams via `@EnumSource`; expected counter count derived from the created stats instead of a hardcoded `8`.*

### Minor style nits in the fix commit (not blocking)

- The nested test enum is named `Stream`, which conceptually shadows `java.util.stream.Stream` — something like `StatStream` reads less ambiguously at the `@EnumSource`/`stream.create(...)` call sites.
- The top-level `INTEGRATION_PROCESSOR` test constant now duplicates `Stream.MESSAGE.key`; the guard test could reference the latter and drop the constant.
- `removeInactiveStats`'s comment documents caller behavior ("one final stats line above") and hard-references `StatsManagerImpl`, both of which can drift — it could be trimmed to just the remap block's own rationale.

---

PR #337, opened 2026-07-03
fix(mqtt): return per-filter UNSUBACK reason codes

dmytro-landiak:
## Pull Request description

Fixes the UNSUBACK reason codes on the MQTT UNSUBSCRIBE path and removes redundant work in `MqttUnsubscribeHandler`.

**Problem**
- Every UNSUBACK reason code was hardcoded to `SUCCESS`. MQTT 5 requires `NO_SUBSCRIPTION_EXISTED (0x11)` for a topic filter the client was not subscribed to.
- `getRemovedSubscriptions()` ran on **every** UNSUBSCRIBE — copying the client's entire subscription set and rebuilding an index — even though that result is only needed when an integration listens for `CLIENT_UNSUBSCRIBED`. The subscriptions actually removed were derived twice, using two different matching rules.

**Change**
- The removal is now the single source of truth. `ClientSubscriptionService.unsubscribeAndPersist(...)` gains an overload that reports the removed subscriptions through a new `UnsubscribeCallback`, collected as a by-product of the existing removal loop (no extra pass, no full-set copy). The existing `BasicCallback`/no-arg overloads are unchanged.
- `MqttUnsubscribeHandler` builds both the per-filter UNSUBACK codes and the `CLIENT_UNSUBSCRIBED` lifecycle events from that single result: a requested filter that matched a removed subscription gets `SUCCESS`, one the client was not subscribed to gets `NO_SUBSCRIPTION_EXISTED`.
- `getRemovedSubscriptions()` and its lookup helper are removed. Added `MqttReasonCodeResolver.unsubAckNoSubscriptionExisted(ctx)`.

**Compatibility**
- MQTT 3.1.1 is unchanged: no reason codes are sent on the wire.
- MQTT 5 clients unsubscribing a filter they were not subscribed to now receive `NO_SUBSCRIPTION_EXISTED (0x11)` instead of `SUCCESS`. Both are success-class reason codes (`< 0x80`), so no client treats this as an error.

**Documentation**
- No documentation changes required.

## General checklist

No front-end changes.

## Back-End feature checklist

- [x] If new dependency was added: the dependency tree is checked for conflicts. _(No new dependencies.)_

### DIFF
--- a/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/handlers/MqttUnsubscribeHandler.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/handlers/MqttUnsubscribeHandler.java
@@ -81,14 +81,10 @@ public class MqttUnsubscribeHandler {
                 .map(TopicSubscription::getTopicFilter)
                 .collect(Collectors.toSet());
         return requestedTopics.stream()
-                .map(topic -> removedTopicFilters.contains(toTopicFilter(topic))
+                .map(topic -> removedTopicFilters.contains(NettyMqttConverter.getTopicFilter(topic))
                         ? MqttReasonCodeResolver.unsubAckSuccess(ctx)
                         : MqttReasonCodeResolver.unsubAckNoSubscriptionExisted(ctx))
                 .collect(Collectors.toList());
-    }
-
-    private static String toTopicFilter(String topic) {
-        return NettyMqttConverter.isSharedTopic(topic) ? NettyMqttConverter.getTopicFilter(topic) : topic;
     }
 
     private void stopProcessingApplicationSharedSubscriptions(ClientSessionCtx ctx, List<String> topics) {
--- a/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/handlers/MqttUnsubscribeHandler.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/handlers/MqttUnsubscribeHandler.java
@@ -50,7 +50,7 @@ public class MqttUnsubscribeHandler {
     public void process(ClientSessionCtx ctx, MqttUnsubscribeMsg msg) {
         log.trace("[{}][{}] Processing unsubscribe, messageId - {}, topic filters - {}", ctx.getClientId(), ctx.getSessionId(), msg.getMessageId(), msg.getTopics());
 
-        clientSubscriptionService.unsubscribeAndPersist(ctx.getClientId(), msg.getTopics(),
+        clientSubscriptionService.unsubscribeAndPersistReportingRemoved(ctx.getClientId(), msg.getTopics(),
                 new UnsubscribeCallback() {
                     @Override
                     public void onSuccess(List<TopicSubscription> removedSubscriptions) {
--- a/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/handlers/MqttUnsubscribeHandler.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/handlers/MqttUnsubscribeHandler.java
@@ -33,6 +33,7 @@ import org.thingsboard.mqtt.broker.service.subscription.shared.TopicSharedSubscr
 import org.thingsboard.mqtt.broker.session.ClientSessionCtx;
 import org.thingsboard.mqtt.broker.util.MqttReasonCodeResolver;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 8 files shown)


application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/subscription/ClientSubscriptionService.java
```diff
@@ -43,6 +43,8 @@ public interface ClientSubscriptionService extends ClientSubscriptionCache {
 
     void unsubscribeAndPersist(String clientId, Collection<String> topicFilters, BasicCallback callback);
 
+    void unsubscribeAndPersist(String clientId, Collection<String> topicFilters, UnsubscribeCallback callback);
```
dmytro-landiak: This adds a second 3-arg `unsubscribeAndPersist` that differs from the `BasicCallback` one only by callback type. Since neither `BasicCallback` nor `UnsubscribeCallback` is a functional interface (both declare two abstract methods), a caller can't disambiguate with a lambda, and a `null` argument becomes an ambiguous-call compile error — the ambiguity already surfaced in the tests, which had to change bare `any()` to `any(BasicCallback.class)`. Worth noting too that the `BasicCallback` 3-arg variant now has no external callers left; it's only reached through the 2-arg delegate. Would distinct names read more clearly — keep `unsubscribeAndPersist(...)` for the fire-and-forget path and add something like `unsubscribeAndPersistReportingRemoved(...)` — rather than leaning on overload resolution by callback type?

application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/subscription/UnsubscribeCallback.java
```diff
@@ -0,0 +1,32 @@
...
+import org.thingsboard.mqtt.broker.common.data.subscription.TopicSubscription;
+
+import java.util.List;
+
+/**
+ * Callback for {@link ClientSubscriptionService#unsubscribeAndPersist(String, java.util.Collection, UnsubscribeCallback)}
+ * that reports the subscriptions actually removed by the operation, so callers can derive the correct
+ * UNSUBACK reason codes and emit lifecycle events without re-deriving what was removed.
+ */
+public interface UnsubscribeCallback {
```
dmytro-landiak, line 27: A dedicated result-carrying callback matches the existing `HttpAuthCallback` precedent, so this isn't wrong — but the codebase now has `BasicCallback`, `HttpAuthCallback`, and this, all identical except the `onSuccess` payload, plus a generic `IntegrationCallback<T>` over in the integration module. Was a generic `Callback<T>` in `common/data` considered to head off further proliferation? Judgment call rather than a defect — the bespoke interface is more self-documenting.

application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/handlers/MqttUnsubscribeHandler.java
```diff
@@ -53,62 +50,47 @@ public class MqttUnsubscribeHandler {
...
+                        if (!removedSubscriptions.isEmpty()) {
+                            integrationLifecycleEventPublisher.publishUnsubscribed(ctx, removedSubscriptions);
+                        }
+                    }
 
-        stopProcessingApplicationSharedSubscriptions(ctx, msg.getTopics());
-    }
+                    @Override
+                    public void onFailure(Throwable t) {
+                        log.warn("[{}][{}] Failed to process client unsubscription", ctx.getClientId(), ctx.getSessionId(), t);
```
dmytro-landiak, line 70: This `onFailure` branch isn't exercised by any test — every test drives `onSuccess`. It's only a warn log, but a case that captures the callback, calls `onFailure`, and asserts no UNSUBACK is written and no lifecycle event is published would lock the behavior in. Related, and pre-existing: on persist failure the client receives no UNSUBACK at all (just this log), so it's left waiting for an ack it never gets — worth confirming that's the intended contract.

dmytro-landiak (commented):
## Review summary

Reviewed 8 changed files in [fix(mqtt): return per-filter UNSUBACK reason codes](https://github.com/thingsboard/tbmq/pull/337). The change is solid — the reason-code fix is correct, MQTT 3.1.1 stays code-free on the wire, and deriving both the codes and the `CLIENT_UNSUBSCRIBED` events from the single removed set delivered through the callback is a clean simplification with good coverage. No functional bugs found; left 6 comments inline — mostly design/quality nits plus two test-coverage gaps.

## Additional findings

- **application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/handlers/MqttUnsubscribeHandler.java:91** — `toTopicFilter` does `isSharedTopic(topic) ? getTopicFilter(topic) : topic`, but `NettyMqttConverter.getTopicFilter` already returns its input unchanged for a non-shared topic, so this is equivalent to just `getTopicFilter(topic)`. `ClientSubscriptionServiceImpl.extractTopicFilterFromSharedTopic` carries the same redundant ternary. Now that `getCodes` matching must stay in lockstep with how `unsubscribe()` strips share prefixes, a single shared helper (delegating to `getTopicFilter`) would remove the risk of the two copies drifting.

---

PR #338, opened 2026-07-03
fix(stats): counter-accuracy fixes for subscription, retained and flow-control metrics

dmytro-landiak:
## Pull Request description

Accuracy fixes for existing broker metrics, found during a metrics audit. Four small, independent fixes to how a handful of stats counters/gauges are recorded — no metric names or descriptions change, and nothing is removed from the public metric surface. Each fix has a focused unit test proving the defect (RED) and the correction (GREEN).

**Back-end scope of changes:**

1. **`clientSubscriptionsConsumer` — record counters only on committed packs.** `logTotal(size)` ran before the classification loop, so a mid-pack failure advanced `totalSubscriptions` while `accepted/ignored` and the offset commit were skipped, breaking the `total == accepted + ignored` invariant and double-counting the monotonic total on a later rebalance re-delivery. The per-pack processing is extracted into a testable method and all three counters are now recorded together only after the whole pack is processed and about to be committed.

2. **`retainedMsgConsumer` — record counters only on committed packs.** Same pattern/fix as (1) for the retained-message consumer.

3. **`flowControl.inflightCount` — guard against duplicate-packetId over-count.** `addInFlightMsg` reserved a slot and called `incInflight()` unconditionally, so re-delivering an already in-flight packetId (e.g. a `DUP=1` retransmission of an unacked persisted QoS1/2 message on channel-writable, whose packetId is cleared from the in-flight set only on ack) bumped the gauge again without growing the set — a drift the ack path never decremented. The duplicate is still reserved and delivered (a retransmission is not a drop), but `incInflight()` now fires only when a genuinely new slot is added; the same guard is applied in the delay-queue drain.

4. **Remove dead no-arg `FlowControlStats.incDropTtl()` overload.** It had no production caller (TTL drops are always counted via `incDropTtl(int)`); removed from the interface and both implementations.

**Investigated, no change needed:** `droppedLifecycleEvents` per-integration counting on a synchronous send failure — verified that the Kafka send is already caught synchronously inside `TbPublishServiceImpl.send` and routed to `droppedEventCallback.onFailure`, so drops are already counted per integration; the outer `onPublishError` correctly counts one drop for pre-fan-out failures (proto build / cache lookup). A per-integration guard was added and then reverted in this branch as unnecessary — the commits are kept for transparency, but the net effect is no change to the lifecycle publisher.

**Documentation:** none required — no metric names, tags, or descriptions change; this only corrects when/how existing series are incremented.

## General checklist

- [x] Description references specific issue. (Internal metrics-audit follow-up; no public issue.)

- [x] N/A — back-end only.

## Back-End feature checklist

- [x] Added corresponding unit and/or integration test(s).
- [x] If new dependency was added: the dependency tree is checked for conflicts. (No new dependencies.)

### DIFF
--- a/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/state/PublishedInFlightCtxImpl.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/state/PublishedInFlightCtxImpl.java
@@ -78,12 +78,17 @@ public class PublishedInFlightCtxImpl implements PublishedInFlightCtx {
 
         boolean reserve = false;
         boolean register = false;
+        boolean newInFlightSlot = false;
         MqttPublishMessage toRelease = null;
 
         lock.lock();
         try {
             if (delayedMsgQueue.isEmpty() && inFlightPacketIds.size() < clientReceiveMax) {
-                inFlightPacketIds.add(packetId);
+                // add() returns false when the packetId is already in-flight (e.g. a DUP retransmission of an
+                // unacked persisted message, whose packetId is cleared only on ack): still deliver it, but only
+                // bump the inflight gauge for a genuinely new slot so it is not over-counted (the ack path
+                // decrements each packetId once).
+                newInFlightSlot = inFlightPacketIds.add(packetId);
                 reserve = true;
             } else if (delayedMsgCounter.get() < delayedMsgQueueMaxSize) {
                 delayedMsgQueue.add(new MqttPubMsgWithCreatedTime(mqttPubMsg, System.nanoTime()));
@@ -97,7 +102,9 @@ public class PublishedInFlightCtxImpl implements PublishedInFlightCtx {
         }
 
         if (reserve) {
-            stats.incInflight();
+            if (newInFlightSlot) {
+                stats.incInflight();
+            }
         } else if (register) {
             log.debug("[{}][{}] Max in-flight messages reached! Buffering msg in delay queue", clientId, clientReceiveMax);
             stats.incDelayed();
@@ -147,6 +154,7 @@ public class PublishedInFlightCtxImpl implements PublishedInFlightCtx {
         while (clientSessionCtx.isWritable()) {
             MqttPublishMessage toSend = null;
             boolean queueEmpty = false;
+            boolean reserved = false;
 
             lock.lock();
             try {
@@ -159,7 +167,9 @@ public class PublishedInFlightCtxImpl implements PublishedInFlightCtx {
                 } else {
                     delayedMsgQueue.poll();
                     delayedMsgCounter.decrementAndGet();
-                    inFlightPacketIds.add(head.getMqttPublishMessage().variableHeader().packetId());
+                    // add() returns false if the drained packetId is already in-flight; only bump the gauge for a
+                    // genuinely new slot so it is not over-counted.
+                    reserved = inFlightPacketIds.add(head.getMqttPublishMessage().variableHeader().packetId());
                     toSend = head.getMqttPublishMessage();
                 }
             } finally {
@@ -172,7 +182,9 @@ public class PublishedInFlightCtxImpl implements PublishedInFlightCtx {
             }
 
             stats.decDelayed();
-            stats.incInflight();
+            if (reserved) {
+                stats.incInflight();
+            }
             deliveryService.sendAlreadyTrackedPublishMsgToClient(clientSessionCtx, toSend);
         }
     }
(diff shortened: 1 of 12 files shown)


application/src/test/java/org/thingsboard/mqtt/broker/service/subscription/ClientSubscriptionConsumerImplTest.java
```diff
@@ -0,0 +1,95 @@
...
+                mock(SubscriptionPersistenceService.class),
+                mock(TbQueueAdmin.class),
+                statsManager);
+    }
+
+    @Test
+    public void givenFullyProcessedPack_whenProcessPack_thenAllCountersAdvanceTogether() {
+        // Two dummy-prefixed records take the ignore fast path (no proto/callback needed) — the point of this test
+        // is that total and the accepted/ignored split are both recorded for a successfully processed pack.
+        List<TbProtoQueueMsg<ClientSubscriptionsProto>> messages = List.of(dummyMsg(), dummyMsg());
```
dmytro-landiak, line 65: This happy-path case feeds two dummy-prefixed records, so it only exercises the `ignoredSubscriptions++` fast path and always asserts `log(0, 2)`. The `accepted` branch — a real (non-dummy) record where `callback.accept(...)` returns true — is never covered, and neither is a real record that returns false. Since the point of the change is that total/accepted/ignored advance together, would it be worth adding a case with a real message so a non-zero `accepted` count and the split are actually asserted (e.g. `log(1, 1)`)? As is, a regression that miscounted accepted subscriptions would still pass this suite.

application/src/test/java/org/thingsboard/mqtt/broker/service/mqtt/retain/RetainedMsgConsumerImplTest.java
```diff
@@ -0,0 +1,95 @@
...
+                mock(RetainedMsgPersistenceService.class),
+                mock(TbQueueAdmin.class),
+                statsManager);
+    }
+
+    @Test
+    public void givenFullyProcessedPack_whenProcessPack_thenAllCountersAdvanceTogether() {
+        // Two empty (tombstone) retained-msg records take the cleared path — the point of this test is that total
+        // and the new/cleared split are both recorded for a successfully processed pack.
+        List<TbProtoQueueMsg<RetainedMsgProto>> messages = List.of(clearedMsg(), clearedMsg());
```
dmytro-landiak, line 66: Same coverage gap as the subscription-consumer test: both records take the cleared (tombstone) path, so only `clearedRetainedMsgCount` is exercised and the assertion is always `log(0, 2)`. The `newRetainedMsgCount++` branch — a non-empty retained message that runs `convertToRetainedMsg` and the header-expiry plumbing — is never hit. Could a mixed pack with a genuine non-empty `RetainedMsgProto` be added so the `new` side of the split is verified too (e.g. `log(1, 1)`)?

application/src/main/java/org/thingsboard/mqtt/broker/actors/client/state/PublishedInFlightCtxImpl.java
```diff
@@ -172,7 +182,9 @@ private void tryDrain() {
             }
 
             stats.decDelayed();
-            stats.incInflight();
+            if (reserved) {
```
dmytro-landiak, line 185: The duplicate-slot guard in `addInFlightMsg` is nicely covered by `addInFlightMsg_duplicatePacketId_...`, but this mirrored branch in `tryDrain` — draining a buffered message whose packetId is already in-flight, so `add()` returns false and `reserved` stays false — has no test. It's reachable: a DUP can arrive while the window is full, get buffered under a packetId still in the in-flight set, then be drained on a later ack. Worth a test that buffers a message with an already-in-flight packetId, acks a different slot, and asserts the drain still delivers it but doesn't call `incInflight()` again?

dmytro-landiak (commented):
## Review summary

Reviewed 10 changed files in [fix(stats): counter-accuracy fixes for subscription, retained and flow-control metrics](https://github.com/thingsboard/tbmq/pull/338). Left 3 comment(s) inline.

The four accuracy fixes read correctly. The `flowControl.inflightCount` accounting is sound — a packetId is `incInflight`'d once on its first `add` and `decInflight`'d once on its first ack, so duplicates no longer double-count while still being reserved and delivered as retransmissions. The consumer `processPack` extraction is behavior-identical apart from the intended `logTotal` timing, and the dead `incDropTtl()` removal has no remaining callers. No correctness issues found. All three inline comments are test-coverage gaps.

---

PR #340, opened 2026-07-06
fix(stats): NaN guards + SQL queue depth gauge + drop dead queueSize log field

dmytro-landiak:
## Pull Request description

Metrics observability hardening for the periodic stats families: guard the average-pack-size log values against `NaN`, export the SQL blocking-queue depth to Prometheus, and drop a dead always-`0` log field. Purely additive/log-hygiene — no existing Prometheus series is renamed or removed.

**Scope of changes**

1. **Guard `getAvgPackSize()` against `NaN` at idle.** `DefaultPublishMsgConsumerStats`, `DefaultDeviceProcessorStats`, and `DefaultClientSessionEventConsumerStats` computed `Math.ceil(totalPackSize / packProcessingTimer.getCount())` with no divisor guard, so an idle print interval (`count == 0`) produced `0/0 → Math.ceil(NaN) → NaN` in the periodic stats log line. They now return `0` when no pack was processed, mirroring `ResettableTimer#getAvg`. These averages are log-only diagnostics (not exported meters); the client-session-event one is documented as intentionally log-only (its pack **processing time** is already exported as a timer).

2. **Export the SQL queue depth as a Micrometer gauge.** The live SQL blocking-queue depth (`queue::size`, wired by `TbSqlBlockingQueue#init`) was computed and printed but never registered as a meter, so this backpressure/durability signal was invisible to Prometheus. `StatsManagerImpl#createSqlQueueStats` now registers a gauge `sqlQueue.<queueName>.queueSize{queueIndex}` over the existing `getCurrentQueueSize` supplier. The backing `MessagesStats` is strong-held in `managedStats`, so the gauge (weak-ref to its state object) is not subject to a GC trap. Additive — new meter name, no existing series changes.

3. **Drop the dead always-`0` `queueSize` log field for producer stats.** `MSG_DISPATCHER_PRODUCER` and `IE_UPLINK_PRODUCER` never wire a queue-size supplier (only the SQL blocking queues do), so their periodic log line always printed a meaningless `queueSize = [0]`. Added `MessagesStats#isQueueSizeTracked()` and consolidated the duplicated `queueSize/totalMsgs/successfulMsgs/failedMsgs` string built by both `StatsManagerImpl#printStats` and `DefaultIntegrationStatisticsService#printStats` into a shared `MessagesStats#toLogString()` default method that emits `queueSize` only when a real depth is tracked. The SQL-queue log lines keep their real depth; producer/ie-uplink lines drop the dead field.

Also adds `spring-boot-starter-test` (test scope) to `common/stats`, which previously had no test infrastructure — mirroring `common/data`.

**Documentation notes:** the new `sqlQueue.<queueName>.queueSize` gauge can be added to the metrics documentation. No configuration changes.

## General checklist

Not applicable — no front-end changes.

## Back-End feature checklist


dmytro-landiak (commented):
## Review summary

Reviewed 14 changed files in [fix(stats): NaN guards + SQL queue depth gauge + drop dead queueSize log field](https://github.com/thingsboard/tbmq/pull/340). Left 4 comment(s) inline.

The correctness pass found no bugs: the NaN guard reads the pack count once and promotes cleanly to `0`, the new SQL-queue gauge establishes its strong reference (`managedStats.add`) before Micrometer's weak-ref registration so there's no GC trap, `(queueName, queueIndex)` is unique per pool so no gauge collides, and the `volatile` supplier closes the newly-added scrape-thread read. The inline comments are all quality/design observations about duplication and where the shared log/avg logic and its rationale should live.

---

common/stats/src/main/java/org/thingsboard/mqtt/broker/common/stats/MessagesStats.java
```diff
@@ -50,4 +50,27 @@ default void incrementFailed() {
...
+     * stats never wire a supplier and would otherwise print a dead always-0 {@code queueSize} log field.
+     */
+    boolean isQueueSizeTracked();
+
+    /**
+     * Renders the counters for the periodic stats log line. The {@code queueSize} field is emitted only
+     * when a real queue depth is tracked ({@link #isQueueSizeTracked()}), so producer-style stats no
+     * longer print a dead always-0 value.
+     */
+    default String toLogString() {
```
dmytro-landiak: `toLogString()` (and the `isQueueSizeTracked()` predicate that exists mainly to feed it) puts a specific log-line format onto `MessagesStats`, which is otherwise a counter recorder/exposer — so `StatsConstantNames` formatting and the "omit queueSize when untracked" rule now live on the interface and every implementer inherits them. The upside is real: it collapses two byte-identical inline formatters (`StatsManagerImpl` and `DefaultIntegrationStatisticsService`). Would a small static helper (e.g. `MessagesStatsLog.format(stats)`) next to the stats-logging code get the same dedup while keeping the interface a pure recorder API? Judgment call, not clear-cut.

application/src/main/java/org/thingsboard/mqtt/broker/service/stats/DefaultPublishMsgConsumerStats.java
```diff
@@ -124,7 +124,8 @@ public double getAvgPackProcessingTime() {
 
     @Override
     public double getAvgPackSize() {
-        return Math.ceil((double) totalPackSize.get() / packProcessingTimer.getCount());
+        int count = packProcessingTimer.getCount();
+        return count == 0 ? 0 : Math.ceil((double) totalPackSize.get() / count);
```
dmytro-landiak: This exact `getAvgPackSize()` body — plus the `totalPackSize` field, its `addAndGet`, the reset, and now the `count == 0` guard — is duplicated verbatim across `DefaultPublishMsgConsumerStats`, `DefaultDeviceProcessorStats`, and `DefaultClientSessionEventConsumerStats`. The duplication predates this PR, but hand-copying the same NaN guard into all three is exactly what tends to drift out of sync later. Since all three already pair a `ResettableTimer` with a `totalPackSize`, would a tiny shared accumulator (wrapping the timer count + running total) be worth extracting so the guard lives in one place?

application/src/main/java/org/thingsboard/mqtt/broker/service/stats/DefaultClientSessionEventConsumerStats.java
```diff
@@ -56,9 +56,16 @@ public double getAvgPackProcessingTime() {
         return packProcessingTimer.getAvg();
     }
 
+    /**
+     * Average processed pack size. This is a log-only diagnostic (printed by {@code StatsManagerImpl.printStats}),
+     * intentionally not exported as a Micrometer meter — mirroring the sibling {@code DefaultPublishMsgConsumerStats}
```
dmytro-landiak: The javadoc explaining the design (returns 0 not NaN at idle, deliberately log-only, mirrors the two siblings) was added only here, yet it describes all three identical implementations and even names the other two — while the siblings got the same code fix with no rationale. Either mirror the note on the siblings, or, if `getAvgPackSize` gets consolidated per the duplication note, move the rationale to wherever the shared code lands. As is, the one place documenting the invariant isn't where a future editor is most likely to look.

application/src/test/java/org/thingsboard/mqtt/broker/service/stats/DefaultPublishMsgConsumerStatsTest.java
```diff
@@ -0,0 +1,62 @@
...
+    private DefaultPublishMsgConsumerStats stats;
+
+    @Before
+    public void setUp() {
+        StatsFactory statsFactory = new DefaultStatsFactory(new SimpleMeterRegistry());
+        stats = new DefaultPublishMsgConsumerStats("consumer-1", statsFactory);
+    }
+
+    @Test
+    public void givenNoPacksProcessed_whenGetAvgPackSize_thenReturnsZeroNotNaN() {
```
dmytro-landiak: `DefaultPublishMsgConsumerStatsTest`, `DefaultDeviceProcessorStatsTest`, and `DefaultClientSessionEventConsumerStatsTest` are byte-for-byte identical apart from the SUT type and the log-pack method name (`logPackProcessingTime` vs `logClientIdPacksProcessingTime`). Three copies of the same three tests have to be edited in lockstep. Since the pack-size behavior is really one shared invariant, would a single parameterized test (or an abstract base with an overridden "log a pack of size N" hook) express it more honestly? The per-class locality is a mild plus, so it's a judgment call.

common/stats/src/main/java/org/thingsboard/mqtt/broker/common/stats/MessagesStats.java
dmytro-landiak: Fixed in ec82842a5 — moved the formatting into a static MessagesStatsLog.format(stats) helper in common/stats (reachable from both print loops it dedups). isQueueSizeTracked() stays on the interface as a state query the formatter reads, not presentation.

application/src/main/java/org/thingsboard/mqtt/broker/service/stats/DefaultPublishMsgConsumerStats.java
dmytro-landiak: Fixed in 98e21877a — extracted a shared PackSizeStats accumulator (common/stats, next to ResettableTimer) that owns the running total + count, the count==0 -> 0 guard, and reset; all three consumer stats now delegate to it. Behavior-identical: record() increments the same total/count the old code did, at the same call site.

application/src/main/java/org/thingsboard/mqtt/broker/service/stats/DefaultClientSessionEventConsumerStats.java
dmytro-landiak: Fixed in 98e21877a — the rationale now lives once, on the shared PackSizeStats primitive, instead of on a single sibling; the per-class javadoc was removed.

application/src/test/java/org/thingsboard/mqtt/broker/service/stats/DefaultPublishMsgConsumerStatsTest.java
dmytro-landiak: Fixed in 98e21877a — the three byte-identical test classes were replaced by a single parameterized ConsumerStatsAvgPackSizeTest (fresh instances per case via a fixture factory), with the guard logic covered directly by PackSizeStatsTest.

common/stats/src/main/java/org/thingsboard/mqtt/broker/common/stats/PackSizeStats.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * <p>This is a log-only diagnostic and is deliberately NOT exported as a Micrometer meter — only the
+ * pack processing <i>time</i> is. {@link #getAvg()} returns 0 (not NaN) at idle intervals where no
+ * pack was recorded, mirroring {@link ResettableTimer#getAvg()}; this matters for sparse streams
+ * (e.g. client-session events split across several consumers) whose intervals are often empty.
+ */
+public class PackSizeStats {
+
+    private final AtomicLong totalPackSize = new AtomicLong();
+    private final AtomicInteger packCount = new AtomicInteger();
```
dmytro-landiak: `PackSizeStats` keeps its own `packCount`, but every caller bumps it in lockstep with the paired `ResettableTimer`'s `logCount` — `packSizeStats.record(packSize)` sits right next to `packProcessingTimer.logTime(...)` in the same method. So there are now two independent counters tracking the identical "packs processed this interval", kept in sync by hand; a future call that records a size without logging a time (or vice versa) would silently skew the average. Was the fully self-contained `PackSizeStats` (which is nice for the standalone `PackSizeStatsTest`) the deliberate choice, or would a combined pack-processing holder that owns both the timer and the size total under one count read cleaner? The timer+size pair is still declared, recorded, and reset together in all three consumer-stats classes.

common/stats/src/main/java/org/thingsboard/mqtt/broker/common/stats/MessagesStatsFormatter.java
```diff
@@ -0,0 +1,41 @@
...
+ * <p>The {@code queueSize} field is emitted only when the stats tracks a real queue depth
+ * ({@link MessagesStats#isQueueSizeTracked()}); producer-style stats (which never wire a supplier)
+ * omit the dead always-0 field.
+ */
+public final class MessagesStatsLog {
+
+    private MessagesStatsLog() {
+    }
+
+    public static String format(MessagesStats stats) {
```
dmytro-landiak, line 31: The class is named `MessagesStatsLog` but it doesn't log anything or hold any log state — it purely formats a `MessagesStats` into a counters string that the caller then passes to `log.info`. The `Log` suffix reads like it writes to a logger; would `MessagesStatsFormatter` set expectations better for the next reader? Minor naming point.

dmytro-landiak (commented):
## Re-review summary

Re-reviewed [fix(stats): NaN guards + SQL queue depth gauge + drop dead queueSize log field](https://github.com/thingsboard/tbmq/pull/340) — verified 4 finding(s) from [previous review](https://github.com/thingsboard/tbmq/pull/340#pullrequestreview-4636056067).

| Status | Count |
|--------|-------|
| ✅ Resolved | 4 |
| 💬 Acknowledged | 0 |
| ❌ Unresolved | 0 |

Also found 2 new (minor, design-level) issue(s) in the fix commits, commented inline.

### Finding details

- ✅ **common/stats/.../MessagesStats.java** — log-line formatting (`toLogString`) baked onto the recorder interface — *Fixed: moved to a static `MessagesStatsLog.format`; `isQueueSizeTracked()` kept on the interface as a state query.*
- ✅ **application/.../DefaultPublishMsgConsumerStats.java** — `getAvgPackSize()` + `totalPackSize` + guard triplicated across the 3 consumer classes — *Fixed: extracted a shared `PackSizeStats`; all three delegate.*
- ✅ **application/.../DefaultClientSessionEventConsumerStats.java** — design-rationale javadoc on only one of the three identical impls — *Fixed: the rationale now lives once, on `PackSizeStats`.*
- ✅ **application/.../DefaultPublishMsgConsumerStatsTest.java** — three byte-identical test classes — *Fixed: consolidated into a parameterized `ConsumerStatsAvgPackSizeTest` + a focused `PackSizeStatsTest`.*

---

common/stats/src/main/java/org/thingsboard/mqtt/broker/common/stats/PackSizeStats.java
dmytro-landiak: Fixed in ff4e94548 — replaced PackSizeStats with PackProcessingStats, which owns the processing-time timer AND the pack-size total, so the count now has a single source (the timer). This restores the original single-count semantics (encapsulated) and also consolidates the timer+size pair that was declared/recorded/reset together in all three consumer-stats classes. No more hand-synced second counter.

common/stats/src/main/java/org/thingsboard/mqtt/broker/common/stats/MessagesStatsFormatter.java
dmytro-landiak: Fixed in 4baa3a2d1 — renamed to MessagesStatsFormatter (class, test, and both callers). It only formats, so the name now matches.

PR #341, opened 2026-07-06
fix(stats): measure delivery on write completion and actors.processing in nanoseconds

dmytro-landiak:
## Pull Request description

Two timer-metric fixes on the `application` module stats surface. Both are **backward compatible** — no metric names change.

**1. Record the `actors.processing` timer in nanoseconds.**
`ContextAwareActor.process` timed each actor message with `StopWatch.getTime()` (milliseconds) and `MsgTypeActorProcessingMetricService` recorded `TimeUnit.MILLISECONDS`, so any sub-millisecond actor work rounded to `0` and the timer's mean/max were meaningless. It now uses `StopWatch.getNanoTime()` and records `TimeUnit.NANOSECONDS`, matching the always-on sibling `clientActor.processing.time`. Micrometer/Prometheus still export the timer in seconds — only the internal resolution improves. The metric stays gated behind `actors.system.processing-metrics.enabled` (default `false`).

**2. Measure the `delivery` timer on write completion instead of enqueue.**
The `delivery` timer bracketed the synchronous return of the Netty `writeAndFlush`/`write` calls, which are asynchronous — so it measured the cost of handing the write off to the event loop, not delivery. It is now recorded from a `ChannelFuture` listener when the write completes successfully (once the PUBLISH bytes are written to the socket), across all three send paths in `DefaultMqttPublishMsgDeliveryService`. For the batched no-flush path the `write()` future completes at the subsequent `flush()`, so the measurement is consistent across paths and now includes batch-wait latency. The listener is attached only when stats are enabled (`stats.enabled`), so no per-message listener allocation or event-loop callback is paid on the hot delivery path when metrics are off. The metric name `delivery` is intentionally kept.

**Operator note — re-baseline `delivery` after upgrade:** the timer now counts only successfully-completed writes (previously every enqueue) and includes batch-wait latency on the batched path, so dashboards/alerts on `delivery` count and latency should be re-checked.

## General checklist

- [x] Description references specific [issue](https://github.com/thingsboard/tbmq/issues). — part of the broader Prometheus metrics-usability review; no single dedicated issue.

- N/A — no front-end changes.

## Back-End feature checklist

- [x] Added corresponding unit and/or integration test(s). New tests: `MsgTypeActorProcessingMetricServiceTest`, `ContextAwareActorTest`, and `DefaultMqttPublishMsgDeliveryServiceTest` (9 tests total) asserting against a `SimpleMeterRegistry`.

### DIFF
--- a/application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/delivery/DefaultMqttPublishMsgDeliveryService.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/delivery/DefaultMqttPublishMsgDeliveryService.java
@@ -15,6 +15,7 @@
  */
 package org.thingsboard.mqtt.broker.service.mqtt.delivery;
 
+import io.netty.channel.ChannelFuture;
 import io.netty.handler.codec.mqtt.MqttPublishMessage;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -25,6 +26,7 @@ import org.thingsboard.mqtt.broker.service.stats.timer.DeliveryTimerStats;
 import org.thingsboard.mqtt.broker.session.ClientSessionCtx;
 
 import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
 
 @Service
 @Slf4j
@@ -32,12 +34,14 @@ public class DefaultMqttPublishMsgDeliveryService implements MqttPublishMsgDeliv
 
     private final TbMessageStatsReportClient tbMessageStatsReportClient;
     private final DeliveryTimerStats deliveryTimerStats;
+    private final boolean statsEnabled;
 
     @Autowired
     public DefaultMqttPublishMsgDeliveryService(TbMessageStatsReportClient tbMessageStatsReportClient,
                                                 StatsManager statsManager) {
         this.tbMessageStatsReportClient = tbMessageStatsReportClient;
         this.deliveryTimerStats = statsManager.getDeliveryTimerStats();
+        this.statsEnabled = statsManager.isEnabled();
     }
 
     @Override
@@ -54,8 +58,8 @@ public class DefaultMqttPublishMsgDeliveryService implements MqttPublishMsgDeliv
     public void sendAlreadyTrackedPublishMsgToClient(ClientSessionCtx ctx, MqttPublishMessage msg) {
         try {
             long startTime = System.nanoTime();
-            ctx.getChannel().writeAndFlush(msg);
-            deliveryTimerStats.logDelivery(startTime, TimeUnit.NANOSECONDS);
+            ChannelFuture future = ctx.getChannel().writeAndFlush(msg);
+            recordDeliveryOnSuccess(future, startTime);
         } catch (Exception e) {
             log.warn("[{}][{}] Failed to send PUBLISH msg to MQTT client", ctx.getClientId(), ctx.getSessionId(), e);
             if (!msg.fixedHeader().isRetain()) {
@@ -64,13 +68,13 @@ public class DefaultMqttPublishMsgDeliveryService implements MqttPublishMsgDeliv
         }
     }
 
-    private void processSendPublish(ClientSessionCtx ctx, MqttPublishMessage msg, Runnable processor) {
+    private void processSendPublish(ClientSessionCtx ctx, MqttPublishMessage msg, Supplier<ChannelFuture> processor) {
         try {
             boolean added = ctx.addInFlightMsg(msg);
             if (added) {
                 long startTime = System.nanoTime();
-                processor.run();
-                deliveryTimerStats.logDelivery(startTime, TimeUnit.NANOSECONDS);
+                ChannelFuture future = processor.get();
+                recordDeliveryOnSuccess(future, startTime);
             }
         } catch (Exception e) {
             log.warn("[{}][{}] Failed to send PUBLISH msg to MQTT client", ctx.getClientId(), ctx.getSessionId(), e);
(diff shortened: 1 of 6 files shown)


application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/delivery/DefaultMqttPublishMsgDeliveryService.java
```diff
@@ -64,13 +68,13 @@ public void sendAlreadyTrackedPublishMsgToClient(ClientSessionCtx ctx, MqttPubli
         }
     }
 
-    private void processSendPublish(ClientSessionCtx ctx, MqttPublishMessage msg, Runnable processor) {
+    private void processSendPublish(ClientSessionCtx ctx, MqttPublishMessage msg, Supplier<ChannelFuture> processor) {
         try {
             boolean added = ctx.addInFlightMsg(msg);
             if (added) {
                 long startTime = System.nanoTime();
```
dmytro-landiak, line 75: `startTime` is captured before every write regardless of `statsEnabled`, but when stats are off `recordDeliveryOnSuccess` returns immediately and the value is discarded. Since `statsEnabled` is a final field known up front, would `long startTime = statsEnabled ? System.nanoTime() : 0L;` be worth it here (and at line 60 in `sendAlreadyTrackedPublishMsgToClient`)? It's a small cost — tens of ns and a well-predicted branch — but the whole point of the guard is to keep the disabled path free of metric work, and this leaves one `nanoTime()` call sitting in front of it (HotSpot won't reliably fold away a call gated on an instance-final field).

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/delivery/DefaultMqttPublishMsgDeliveryService.java
```diff
@@ -54,8 +58,8 @@ public void sendPublishMsgToClientWithoutFlush(ClientSessionCtx ctx, MqttPublish
     public void sendAlreadyTrackedPublishMsgToClient(ClientSessionCtx ctx, MqttPublishMessage msg) {
         try {
             long startTime = System.nanoTime();
-            ctx.getChannel().writeAndFlush(msg);
-            deliveryTimerStats.logDelivery(startTime, TimeUnit.NANOSECONDS);
+            ChannelFuture future = ctx.getChannel().writeAndFlush(msg);
+            recordDeliveryOnSuccess(future, startTime);
         } catch (Exception e) {
             log.warn("[{}][{}] Failed to send PUBLISH msg to MQTT client", ctx.getClientId(), ctx.getSessionId(), e);
```
dmytro-landiak, line 64: The two send paths now carry byte-for-byte identical catch blocks (warn + conditional `reportDroppedMsgs`) — here and in `processSendPublish` at line 80. It predates this PR, but since both methods were reworked here, this might be a natural moment to fold the failure handling into a small helper like `reportSendFailure(ctx, msg, e)`. Only two call sites, so it reads fine as-is — optional cleanup rather than a must-fix.

application/src/test/java/org/thingsboard/mqtt/broker/service/mqtt/delivery/DefaultMqttPublishMsgDeliveryServiceTest.java
```diff
@@ -0,0 +1,192 @@
...
+        DefaultMqttPublishMsgDeliveryService service = buildService(true);
+        when(ctx.addInFlightMsg(msg)).thenReturn(true);
+        when(channel.writeAndFlush(msg)).thenReturn(future);
+        when(future.isSuccess()).thenReturn(true);
+
+        service.sendPublishMsgToClient(ctx, msg);
+
+        // Delivery must be recorded when the async write completes, not synchronously at submission.
+        assertEquals(0, deliveryTimer().count());
+        captureListener().operationComplete(future);
```
dmytro-landiak, line 103: These tests drive completion by mocking `ChannelFuture` and invoking the captured listener by hand, which cleanly proves the listener is wired and only records on `isSuccess()`. It doesn't exercise a real Netty write firing the listener, though, so the core assumption — that the future completes at socket-write time, including the batched `flush()` for the no-flush path — rests on Netty's contract rather than on the test. Was an `EmbeddedChannel`-based test considered to lock that in end to end? It's more setup and the mock already covers the branch logic, so this may be a deliberate trade-off worth acknowledging rather than a must-fix.

application/src/test/java/org/thingsboard/mqtt/broker/service/mqtt/delivery/DefaultMqttPublishMsgDeliveryServiceTest.java
```diff
@@ -0,0 +1,192 @@
...
+        DefaultMqttPublishMsgDeliveryService service = buildService(true);
+        MqttFixedHeader fixedHeader = mock(MqttFixedHeader.class);
+        when(fixedHeader.isRetain()).thenReturn(false);
+        when(msg.fixedHeader()).thenReturn(fixedHeader);
+        when(ctx.addInFlightMsg(msg)).thenReturn(true);
+        when(channel.writeAndFlush(msg)).thenThrow(new RuntimeException("boom"));
+
+        service.sendPublishMsgToClient(ctx, msg);
+
+        verify(reportClient, times(1)).reportDroppedMsgs();
```
dmytro-landiak, line 188: This covers the non-retained synchronous-throw branch that reports a drop, but the sibling branch — a retained message that throws, which must NOT report a drop — isn't asserted, nor is the equivalent catch path in `sendAlreadyTrackedPublishMsgToClient`. That logic is pre-existing and unchanged, so this is a light coverage note rather than a gap in the PR's new behavior, but a retained-throw assertion would pin the `if (!isRetain())` guard against future edits to these reworked methods.

application/src/test/java/org/thingsboard/mqtt/broker/actors/service/ContextAwareActorTest.java
```diff
@@ -0,0 +1,60 @@
...
+        ActorProcessingMetricService capturingService = (msgType, time) -> captured.set(time);
+
+        ActorSystemContext systemContext = mock(ActorSystemContext.class);
+        when(systemContext.getActorProcessingMetricService()).thenReturn(capturingService);
+
+        ContextAwareActor actor = new ContextAwareActor(systemContext) {
+            @Override
+            protected boolean doProcess(TbActorMsg msg) {
+                try {
+                    Thread.sleep(2);
```
dmytro-landiak, line 43: Distinguishing ns from ms via a real 2 ms sleep against a 100_000 ns floor works, and the ~20x margin makes false failures very unlikely (the sleep never returns early, and a millisecond reading of ~2 stays far below the floor), so I don't think this is meaningfully flaky. It does couple the test to wall-clock time and adds ~2 ms of runtime, though. `StopWatch` is constructed internally so it can't be injected, which is presumably why the sleep is used; if it ever became injectable, a controllable clock would make this instant and deterministic. Fine as-is — just flagging the trade-off.

dmytro-landiak (commented):
## Review summary

Reviewed 6 changed files in [fix(stats): measure delivery on write completion and actors.processing in nanoseconds](https://github.com/thingsboard/tbmq/pull/341). Left 5 comments inline.

Both fixes look correct and non-breaking (no metric-name or external-API change): the `actors.processing` unit switch is isolated — the only other `ActorProcessingMetricService` impl is a no-op stub, and the sibling `clientActor.processing.time` already records nanoseconds — and the `delivery` listener records on success only, is guarded on `stats.enabled`, and introduces no listener leak. The inline comments are all optional quality/test-coverage notes: a hot-path `nanoTime()` micro-optimization, duplicated catch blocks, and a few test-design/coverage trade-offs.

---

PR #343, opened 2026-07-06
fix(stats): consistent 'subscriptions' gauge, drop dead processedBytes key, rolling-upgrade guard

dmytro-landiak:
## Pull Request description

Part of the ongoing TBMQ metrics cleanup, focused on making the Prometheus meters consistent with the historical (UI-monitoring) timeseries. Three related changes, plus a rolling-upgrade safety fix surfaced during review.

### 1. `clientSubscriptions` gauge → `subscriptions` (renamed **and** redefined) — breaking

The Micrometer gauge `clientSubscriptions` was registered as `Map::size` over the client→subscriptions map, so it reported the **number of clients that have at least one subscription** — not a subscription count. Its name was near-identical to the historical `subscriptions` chart, which measures **total subscriptions**, so the two silently disagreed for any client with more than one subscription.

- The gauge is renamed to `subscriptions` and redefined to the **total subscription count** (sum of the per-client subscription-set sizes), computed consistently with the historical chart's `ClientSubscriptionService#getClientSubscriptionsCount()`. Prometheus and the UI `subscriptions` chart now report the same value from the same source.
- Renaming (instead of silently changing the value under the same name) makes the change fail loud — old `clientSubscriptions` queries return no data rather than quietly reporting a different quantity.

**Migration:** update dashboards/alerts from `clientSubscriptions` to `subscriptions`. The value also changes: `subscriptions` (total) is `>=` the old `clientSubscriptions` (clients-with-subscriptions) whenever any client holds more than one subscription. Cluster behaviour is unchanged — the backing map is cluster-wide replicated, so each node reports the same cluster-wide total (take one node / `max`, don't sum across nodes).

### 2. Remove the dead `processedBytes` historical key

`processedBytes` was never incremented — it was only ever persisted and queried as `0`, writing an empty timeseries row every reporting interval. It is removed from `MSG_RELATED_HISTORICAL_KEYS` (per-interval persistence) and `HISTORICAL_KEYS` (latest-value query); `MSG_RELATED_HISTORICAL_KEYS_COUNT` is derived from the list, so the report client's latch/list sizing adjusts automatically. The constant is kept until a follow-up upgrade script purges the existing `processedBytes` rows from PostgreSQL. No meaningful series existed, so this is not breaking.

### 3. Rolling-upgrade safety: ignore unknown historical keys

The cluster-wide historical total consumer seeds its aggregation map only with the current key set, but looks it up by the **incoming message's** key and immediately dereferences the result. During a rolling upgrade a not-yet-upgraded node still produces `processedBytes` messages; an upgraded consumer would resolve `null` → NPE. Because `commitSync()` runs only after the per-message loop, the uncommitted batch would be re-polled and re-fail forever — a poison pill stalling all cluster-wide historical aggregation until topic retention drops the message. `processSaveHistoricalStatsTotal` now skips any key absent from the map.

### 4. Consistency clean-up (no behaviour change)

- `ClientSubscriptionServiceImpl#init` folded from two passes into one (map keys are unique by client id, so populate + subscribe + shared-cache-put run safely in a single loop) and now logs both the client count and the total subscription count.
- Type-level javadoc on `TbMessageStatsReportClient` documenting that the historical charts are per-interval deltas (gated on `historical-data-report.enabled`) whereas the Prometheus meters are cumulative/unconditional, plus the known async-save-loss limitation (a failed interval save is lost from the chart but not from Prometheus).
- `StatsType` comments disambiguating `connectedSessions` (live channels on this node), `allClientSessions` (cluster-wide incl. offline persistent sessions — the analog of the historical `sessions` chart) and `subscriptions`.

### Documentation notes

The 2.4 release notes / metrics docs must call out the `clientSubscriptions` → `subscriptions` rename and value redefinition (old → new name; value now = total subscriptions instead of clients-with-subscriptions).

## General checklist

- [x] Description references specific issue.
- [x] Changes are backward compatible or upgrade script is provided. — intentional breaking metric-name/semantics change for 2.4 (see §1); documented above for release notes. `processedBytes` psql cleanup deferred to a follow-up upgrade script.

## Back-End feature checklist

- [x] Added corresponding unit and/or integration test(s).
- [x] If new dependency was added: the dependency tree is checked for conflicts. — no new dependencies.

### DIFF
--- a/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/subscription/ClientSubscriptionServiceImpl.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/subscription/ClientSubscriptionServiceImpl.java
@@ -41,6 +41,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.LongAdder;
 import java.util.stream.Collectors;
 
 import static org.thingsboard.mqtt.broker.common.data.util.CallbackUtil.createCallback;
@@ -62,6 +63,10 @@ public class ClientSubscriptionServiceImpl implements ClientSubscriptionService 
     private final ClientMqttActorManager clientMqttActorManager;
 
     private ConcurrentMap<String, Set<TopicSubscription>> clientSubscriptionsMap;
+    // Running total of subscriptions across all clients, maintained on every subscribe/unsubscribe/clear so
+    // the 'subscriptions' gauge and getClientSubscriptionsCount stay O(1) rather than summing every client's
+    // set on each read (this map is the largest in the broker at the 100M-connection scale it targets).
+    private final LongAdder subscriptionCount = new LongAdder();
     private volatile boolean initialized = false;
 
     @Override
@@ -74,8 +79,9 @@ public class ClientSubscriptionServiceImpl implements ClientSubscriptionService 
             clientSubscriptionsMap.put(clientId, topicSubscriptions);
             subscriptionService.subscribe(clientId, topicSubscriptions);
             sharedSubscriptionCacheService.put(clientId, topicSubscriptions);
+            subscriptionCount.add(topicSubscriptions.size());
         });
-        statsManager.registerSubscriptionsStats(clientSubscriptionsMap);
+        statsManager.registerSubscriptionsStats(subscriptionCount);
         initialized = true;
         log.info("Subscriptions initialized. Clients with subscriptions: {}, total subscriptions: {}",
                 clientSubscriptionsMap.size(), getClientSubscriptionsCount());
@@ -117,6 +123,7 @@ public class ClientSubscriptionServiceImpl implements ClientSubscriptionService 
         subscriptionService.subscribe(clientId, topicSubscriptions);
 
         Set<TopicSubscription> clientSubscriptions = clientSubscriptionsMap.computeIfAbsent(clientId, s -> new HashSet<>());
+        int sizeBefore = clientSubscriptions.size();
         clientSubscriptions.removeIf(sub -> {
             boolean existSubs = topicSubscriptions.contains(sub);
             if (existSubs && sub.isSharedSubscription()) {
@@ -125,6 +132,8 @@ public class ClientSubscriptionServiceImpl implements ClientSubscriptionService 
             return existSubs;
         });
         clientSubscriptions.addAll(topicSubscriptions);
+        // Net change to this client's set == net change to the total (re-subscribes cancel out).
+        subscriptionCount.add(clientSubscriptions.size() - sizeBefore);
         sharedSubscriptionCacheService.put(clientId, topicSubscriptions);
         return clientSubscriptions;
     }
@@ -176,6 +185,7 @@ public class ClientSubscriptionServiceImpl implements ClientSubscriptionService 
             }
             return unsubscribe;
         });
+        subscriptionCount.add(-removedSubscriptions.size());
         return new UnsubscribeResult(clientSubscriptions, removedSubscriptions);
     }
 
@@ -217,6 +227,7 @@ public class ClientSubscriptionServiceImpl implements ClientSubscriptionService 
             log.debug("[{}] There were no active subscriptions for client.", clientId);
             return;
         }
+        subscriptionCount.add(-clientSubscriptions.size());
         List<String> unsubscribeTopics = clientSubscriptions.stream()
                 .peek(topicSubscription -> processSharedUnsubscribe(clientId, topicSubscription))
                 .map(TopicSubscription::getTopicFilter)
@@ -226,7 +237,7 @@ public class ClientSubscriptionServiceImpl implements ClientSubscriptionService 
 
     @Override
     public int getClientSubscriptionsCount() {
-        return clientSubscriptionsMap == null ? 0 : clientSubscriptionsMap.values().stream().mapToInt(Set::size).sum();
+        return subscriptionCount.intValue();
     }
 
     @Override
--- a/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/subscription/ClientSubscriptionServiceImpl.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/subscription/ClientSubscriptionServiceImpl.java
@@ -236,8 +236,8 @@ public class ClientSubscriptionServiceImpl implements ClientSubscriptionService 
     }
 
     @Override
-    public int getClientSubscriptionsCount() {
-        return subscriptionCount.intValue();
+    public long getClientSubscriptionsCount() {
+        return subscriptionCount.sum();
     }
 
     @Override
--- a/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/subscription/ClientSubscriptionServiceImpl.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/subscription/ClientSubscriptionServiceImpl.java
@@ -67,15 +67,18 @@ public class ClientSubscriptionServiceImpl implements ClientSubscriptionService 
     @Override
     public void init(Map<SubscriptionsSourceKey, Set<TopicSubscription>> clientTopicSubscriptions) {
         clientSubscriptionsMap = new ConcurrentHashMap<>();
-        clientTopicSubscriptions.forEach((key, val) -> clientSubscriptionsMap.put(key.getId(), val));
-        statsManager.registerClientSubscriptionsStats(clientSubscriptionsMap);
-
-        clientSubscriptionsMap.forEach((clientId, topicSubscriptions) -> {
+        // Keys are unique by client id (SubscriptionsSourceKey equals/hashCode exclude 'source'),
+        // so a single pass safely populates the map and subscribes each client exactly once.
+        clientTopicSubscriptions.forEach((key, topicSubscriptions) -> {
+            String clientId = key.getId();
+            clientSubscriptionsMap.put(clientId, topicSubscriptions);
             subscriptionService.subscribe(clientId, topicSubscriptions);
             sharedSubscriptionCacheService.put(clientId, topicSubscriptions);
         });
+        statsManager.registerSubscriptionsStats(clientSubscriptionsMap);
         initialized = true;
-        log.info("Subscriptions initialized. Total clients with subscriptions: {}", clientSubscriptionsMap.size());
+        log.info("Subscriptions initialized. Clients with subscriptions: {}, total subscriptions: {}",
+                clientSubscriptionsMap.size(), getClientSubscriptionsCount());
     }
 
     @Override
--- a/application/src/main/java/org/thingsboard/mqtt/broker/service/historical/stats/HistoricalStatsTotalConsumer.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/service/historical/stats/HistoricalStatsTotalConsumer.java
@@ -154,9 +154,9 @@ public class HistoricalStatsTotalConsumer {
                 throwable -> log.error("[{}] Failed to save timeseries entries {}", ENTITY_ID_TOTAL, entries, throwable));
     }
 
-    // Protected (not private) so it can be unit-tested directly, consistent with the other test seams
-    // in this class (calculatePairUsingProvidedMsg, getTotalMessageCounterPair); not an extension point.
-    protected void processSaveHistoricalStatsTotal(TbProtoQueueMsg<ToUsageStatsMsgProto> msg) {
+    // Package-private (not private) so the same-package unit test can invoke it directly, like the other
+    // test seams in this class (calculatePairUsingProvidedMsg, getTotalMessageCounterPair); not an extension point.
+    void processSaveHistoricalStatsTotal(TbProtoQueueMsg<ToUsageStatsMsgProto> msg) {
         String key = msg.getValue().getUsageStats().getKey();
         if (!totalStatsMap.containsKey(key)) {
             // Unknown/removed historical key (e.g. a 'processedBytes' message from a not-yet-upgraded node
@@ -189,7 +189,7 @@ public class HistoricalStatsTotalConsumer {
                 throwable -> log.error("[{}] Failed to save timeseries for key {} with value {}", ENTITY_ID_TOTAL, tsKvEntry.getKey(), tsKvEntry.getValue(), throwable));
     }
 
-    protected TsMsgTotalPair calculatePairUsingProvidedMsg(TbProtoQueueMsg<ToUsageStatsMsgProto> msg) {
+    TsMsgTotalPair calculatePairUsingProvidedMsg(TbProtoQueueMsg<ToUsageStatsMsgProto> msg) {
         TsMsgTotalPair pair = getTotalMessageCounterPair(msg);
         long msgTs = msg.getValue().getTs();
         if (pair.getTs() < msgTs) {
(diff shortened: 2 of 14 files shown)


application/src/main/java/org/thingsboard/mqtt/broker/service/stats/StatsManagerImpl.java
```diff
@@ -304,10 +305,17 @@ public void registerAllClientSessionsStats(Map<?, ?> clientSessionsMap) {
...
+        log.trace("Registering SubscriptionsStats");
+        // Total subscription count (sum of the per-client subscription set sizes), consistent with the
+        // historical 'subscriptions' chart (ClientSubscriptionService#getClientSubscriptionsCount) --
+        // NOT the number of clients that have subscriptions.
+        statsFactory.createGauge(StatsType.SUBSCRIPTIONS.getPrintName(), clientSubscriptionsMap, StatsManagerImpl::countSubscriptions);
+        gauges.add(new Gauge(StatsType.SUBSCRIPTIONS.getPrintName(), () -> countSubscriptions(clientSubscriptionsMap)));
+    }
+
+    private static int countSubscriptions(Map<?, ? extends Collection<?>> clientSubscriptionsMap) {
+        return clientSubscriptionsMap.values().stream().mapToInt(Collection::size).sum();
```
dmytro-landiak: This changes the gauge from O(1) `Map::size` to an O(N) walk over every client's subscription set, and it runs on both read paths — every Prometheus scrape (via `createGauge`) and every `stats.print-interval-ms` log flush (via the internal `Gauge` supplier). Since `clientSubscriptionsMap` is keyed per client and this broker targets ~100M connections, that's a full-map traversal a couple of times a minute on the scrape/log thread. The accurate total is the whole point here, so `Map::size` genuinely can't come back — but did you weigh maintaining a running `LongAdder` updated on the subscribe/unsubscribe write path so the read stays O(1)? The trade-off is a little write-path bookkeeping vs. potentially expensive scrapes; worth a conscious call at this scale.

application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/subscription/ClientSubscriptionServiceImpl.java
```diff
@@ -67,15 +67,18 @@ public class ClientSubscriptionServiceImpl implements ClientSubscriptionService
...
+            String clientId = key.getId();
+            clientSubscriptionsMap.put(clientId, topicSubscriptions);
             subscriptionService.subscribe(clientId, topicSubscriptions);
             sharedSubscriptionCacheService.put(clientId, topicSubscriptions);
         });
+        statsManager.registerSubscriptionsStats(clientSubscriptionsMap);
         initialized = true;
-        log.info("Subscriptions initialized. Total clients with subscriptions: {}", clientSubscriptionsMap.size());
+        log.info("Subscriptions initialized. Clients with subscriptions: {}, total subscriptions: {}",
+                clientSubscriptionsMap.size(), getClientSubscriptionsCount());
```
dmytro-landiak, line 87: This log line walks the whole map a second time via `getClientSubscriptionsCount()`, right after the `forEach` above already iterated it. It's init-only so the cost is one-time and harmless — but you could accumulate the total in a local counter inside the single-pass loop and log that, avoiding the extra traversal.

application/src/test/java/org/thingsboard/mqtt/broker/service/historical/stats/HistoricalStatsTotalConsumerTest.java
```diff
@@ -261,6 +263,19 @@ public void givenStatsMsgs_whenNextTsComes_thenCalculateNewTotalForEachKey() {
...
+    public void givenMsgWithUnknownRemovedKey_whenProcessSaveHistoricalStatsTotal_thenIgnoredWithoutSaveOrNpe() {
+        // processedBytes was removed from the historical key set, so totalStatsMap has no entry for it.
+        // A message with such a key (e.g. produced by a not-yet-upgraded node during a rolling upgrade)
+        // must be ignored, not throw NPE — otherwise the uncommitted batch is re-polled forever and
+        // poisons cluster-wide historical aggregation.
+        var msg = buildMessage(PROCESSED_BYTES, 7);
+
+        historicalStatsTotalConsumer.processSaveHistoricalStatsTotal(msg);
+
+        verify(timeseriesService, never()).save(any(), any(TsKvEntry.class));
```
dmytro-landiak, line 276: This is the first test to drive `processSaveHistoricalStatsTotal`, and it only covers the new negative branch (unknown key → no save). Worth a positive counterpart that feeds a known key (e.g. `INCOMING_MSGS`) and asserts exactly one `save` — otherwise a future change that over-broadens the guard and starts dropping valid keys wouldn't be caught by the suite.

application/src/main/java/org/thingsboard/mqtt/broker/service/historical/stats/HistoricalStatsTotalConsumer.java
```diff
@@ -154,8 +154,15 @@ private void processCountStats() {
                 throwable -> log.error("[{}] Failed to save timeseries entries {}", ENTITY_ID_TOTAL, entries, throwable));
     }
 
-    private void processSaveHistoricalStatsTotal(TbProtoQueueMsg<ToUsageStatsMsgProto> msg) {
+    protected void processSaveHistoricalStatsTotal(TbProtoQueueMsg<ToUsageStatsMsgProto> msg) {
```
dmytro-landiak: Visibility was widened to `protected` only so the test can invoke this. That's consistent with the other test seams in this class (`calculatePairUsingProvidedMsg`, `getTotalMessageCounterPair`), but none of them are marked, so the intent isn't obvious to a future reader who might treat it as a real extension point. A `@VisibleForTesting` annotation (ideally on all of these seams) would make the reason explicit.

dmytro-landiak (commented):
## Review summary

Reviewed 13 changed files in [fix(stats): consistent 'subscriptions' gauge, drop dead processedBytes key, rolling-upgrade guard](https://github.com/thingsboard/tbmq/pull/343). Left 4 comment(s) inline.

The correctness pass found no bugs: the rename is fully propagated (both `StatsManager` implementors + the sole caller updated, no reliance on the old enum constant name), the rolling-upgrade guard runs on a single-threaded executor and can't false-skip a valid key, and the `processedBytes` removal is safe because the derived `MSG_RELATED_HISTORICAL_KEYS_COUNT` keeps the report client's latch/list sizing correct. The inline comments are quality/design observations — most notably the O(N)-per-scrape cost of the redefined `subscriptions` gauge at the scale this broker targets.

## Additional findings

This observation is about existing code outside the PR's diff — spotted while reading surrounding context.

- **application/src/main/java/org/thingsboard/mqtt/broker/service/stats/StatsManagerImpl.java** — the new `countSubscriptions(Map<?, ? extends Collection<?>>)` is a strict generalization of the existing private `getSum(Map<String, List<ApplicationSharedSubscriptionJob>>)`: both sum the sizes of a map's value collections. Now that the general version exists, `registerActiveSharedApplicationProcessorsStats` could reuse it and `getSum` be removed — worth folding the two into one helper to avoid two near-identical private sums in the same class.

---

application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/subscription/ClientSubscriptionServiceImpl.java
```diff
@@ -62,20 +63,28 @@ public class ClientSubscriptionServiceImpl implements ClientSubscriptionService
     private final ClientMqttActorManager clientMqttActorManager;
 
     private ConcurrentMap<String, Set<TopicSubscription>> clientSubscriptionsMap;
+    // Running total of subscriptions across all clients, maintained on every subscribe/unsubscribe/clear so
+    // the 'subscriptions' gauge and getClientSubscriptionsCount stay O(1) rather than summing every client's
+    // set on each read (this map is the largest in the broker at the 100M-connection scale it targets).
+    private final LongAdder subscriptionCount = new LongAdder();
```
dmytro-landiak, line 69: The running total is a manually-maintained shadow of `clientSubscriptionsMap` — the invariant `subscriptionCount == sum of all set sizes` holds only by convention across the four mutation paths. It's correct today, but two things make it easy to break silently later: `getAllClientSubscriptions()` returns `new HashMap<>(clientSubscriptionsMap)`, a shallow copy that hands out the *same* `Set` instances (the current sole caller only reads, so it's fine for now), and any future 5th mutation path has to remember to touch the adder. The field comment helps; a short note on `getAllClientSubscriptions()` that its returned sets must not be mutated would nail down the other half of the invariant.

application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/subscription/ClientSubscriptionServiceImpl.java
```diff
@@ -122,6 +132,8 @@ private Set<TopicSubscription> subscribe(String clientId, Collection<TopicSubscr
             return existSubs;
         });
         clientSubscriptions.addAll(topicSubscriptions);
+        // Net change to this client's set == net change to the total (re-subscribes cancel out).
+        subscriptionCount.add(clientSubscriptions.size() - sizeBefore);
```
dmytro-landiak, line 136: The new drift-guard tests all add distinct new topics, so the subtlest branch of this arithmetic — re-subscribing an already-present (topicFilter+shareName-equal) subscription, where `size() - sizeBefore` must net to zero — isn't asserted against `getClientSubscriptionsCount()`. The shared-sub tests show the map size holding at 1 on a duplicate subscribe, but they read `getClientSubscriptions()`, not the counter. A test that re-subscribes an existing entry and asserts the total is unchanged would cover exactly the line this comment flags as tricky.

application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/subscription/ClientSubscriptionServiceImpl.java
```diff
@@ -223,7 +237,7 @@ private void clearSubscriptions(String clientId) {
 
     @Override
     public int getClientSubscriptionsCount() {
-        return clientSubscriptionsMap == null ? 0 : clientSubscriptionsMap.values().stream().mapToInt(Set::size).sum();
+        return subscriptionCount.intValue();
```
dmytro-landiak: Minor consistency question, not a bug: the backing counter is a `LongAdder`, the gauge reads it via `sum()` (long), and the historical consumer stores the result in a `long`, yet this getter narrows through `intValue()`. Was `int` deliberate to match the pre-existing `ClientSubscriptionService` signature, or would widening the interface to `long` be cleaner now that the whole path is long — especially given the field comment cites the 100M-connection scale as the design driver?

application/src/main/java/org/thingsboard/mqtt/broker/service/historical/stats/HistoricalStatsTotalConsumer.java
```diff
@@ -154,8 +154,17 @@ private void processCountStats() {
                 throwable -> log.error("[{}] Failed to save timeseries entries {}", ENTITY_ID_TOTAL, entries, throwable));
     }
 
-    private void processSaveHistoricalStatsTotal(TbProtoQueueMsg<ToUsageStatsMsgProto> msg) {
+    // Protected (not private) so it can be unit-tested directly, consistent with the other test seams
```
dmytro-landiak: Small one: the comment says this is purely a test seam and "not an extension point," but `protected` is a touch broader than that goal needs — the test is in the same package, so package-private would both suffice and express "test-only, not for subclasses" more precisely. I see the appeal of matching the two sibling `protected` seams for consistency; just flagging that the consistency is propagating a slightly-too-wide visibility. Minor either way.

dmytro-landiak (commented):
## Re-review summary

Re-reviewed [fix(stats): consistent 'subscriptions' gauge, drop dead processedBytes key, rolling-upgrade guard](https://github.com/thingsboard/tbmq/pull/343) — verified 5 finding(s) from [previous review](https://github.com/thingsboard/tbmq/pull/343#pullrequestreview-4642555386).

| Status | Count |
|--------|-------|
| ✅ Resolved | 5 |
| 💬 Acknowledged | 0 |
| ❌ Unresolved | 0 |

All previous findings have been addressed. Also found 4 new observations in the fix commit — all quality/design, not bugs — commented inline. The correctness pass confirmed the new O(1) running counter is maintained consistently across every mutation path (`init`/`subscribe`/`unsubscribe`/`clearSubscriptions`), `init` is single-call (so resetting the map without resetting the `final` counter is safe), and the sole `getAllClientSubscriptions` cache consumer is read-only — so no drift.

### Finding details

- ✅ **StatsManagerImpl** — O(N)-per-scrape `subscriptions` gauge — *Now O(1) via `LongAdder::sum`, with the total maintained on the write path exactly as suggested.*
- ✅ **ClientSubscriptionServiceImpl#init** — log line double-walked the map — *`getClientSubscriptionsCount()` is now O(1), so the read no longer traverses.*
- ✅ **HistoricalStatsTotalConsumerTest** — only the unknown-key branch was tested — *Added a positive known-key test asserting exactly one save.*
- ✅ **HistoricalStatsTotalConsumer** — `protected` seam intent unclear — *Explanatory comment added (comment chosen over `@VisibleForTesting`, which has no precedent in this codebase — a sound call).*
- ✅ **StatsManagerImpl** — `countSubscriptions`/`getSum` near-duplication — *`countSubscriptions` removed entirely.*

---

application/src/main/java/org/thingsboard/mqtt/broker/service/stats/StatsManagerImpl.java
dmytro-landiak: Done in 55411c7d4 — went with the running `LongAdder`. It's seeded in `init`, incremented on `subscribe` by the net set-size delta (`clientSubscriptions.size() - sizeBefore`, so re-subscribes cancel out), and decremented on `unsubscribe`/`clearSubscriptions`. Both the gauge (`LongAdder::sum`) and `getClientSubscriptionsCount()` are now O(1). Confirmed every subscription-set mutation flows through those four paths and the sole `getAllClientSubscriptions` consumer is read-only, so the counter can't drift.

application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/subscription/ClientSubscriptionServiceImpl.java
dmytro-landiak: Addressed in 55411c7d4 — rather than a local counter, `getClientSubscriptionsCount()` now reads the O(1) `LongAdder`, so this log line no longer traverses the map at all.

application/src/test/java/org/thingsboard/mqtt/broker/service/historical/stats/HistoricalStatsTotalConsumerTest.java
dmytro-landiak: Added in 55411c7d4 — `givenMsgWithKnownKey_whenProcessSaveHistoricalStatsTotal_thenSavesExactlyOnce` feeds a known key (`INCOMING_MSGS`) and asserts exactly one `save`, so a future change that over-broadens the guard and starts dropping valid keys would fail the suite.

application/src/main/java/org/thingsboard/mqtt/broker/service/historical/stats/HistoricalStatsTotalConsumer.java
dmytro-landiak: Added an explanatory comment in 55411c7d4 making the test-only intent explicit. I skipped `@VisibleForTesting` deliberately — it has zero usages anywhere in the codebase, so introducing it here would set a new precedent rather than follow one; the comment stays consistent with the surrounding code while documenting the reason.

PR #344, opened 2026-07-08
feat(stats): add clientDisconnects Prometheus counter

dmytro-landiak:
## Pull Request description

Adds a broker-wide **`clientDisconnects`** Micrometer counter (Prometheus `clientDisconnects_total`), giving operators a disconnect-rate signal that CE previously lacked — until now the only session signal was the `connectedSessions` current-count gauge. `rate(clientDisconnects_total[5m])` / `increase(...)` yields disconnects over time, separating churn spikes from steady state.

**Scope of changes**

- New `ClientDisconnectStats` seam in `service.stats` (interface + `DefaultClientDisconnectStats` + no-op `StubClientDisconnectStats`), exposed via `StatsManager.getClientDisconnectStats()`. Mirrors the existing `DroppedLifecycleEventStats` / `DroppedMsgStats` family and rides the existing `stats.enabled` master switch (real impl when enabled, no-op stub otherwise) — no dedicated flag.
- `DisconnectServiceImpl.disconnect` increments the counter for every genuine disconnect of an **established** session — all client- and broker-initiated reasons, including cross-node session takeover (`ON_CLUSTER_CONFLICTING_SESSIONS`). The stats instance is resolved once in `@PostConstruct`, matching the other hot-path consumers (`TbMessageStatsReportClientImpl`, `IntegrationLifecycleEventPublisherImpl`).
- The counter is intentionally **untagged** (a single aggregate series). `DisconnectReasonType` is threaded through `ClientDisconnectStats.increment(reasonType)` as an extension point, but this CE implementation records one untagged counter and ignores the reason.
- The increment fires under the exact same condition as the `CLIENT_DISCONNECTED` lifecycle event, so it excludes:
  - `ON_CONNECTION_FAILURE` — a broker-refused connection that never established a session (a connection *refusal*, not a disconnect);
  - not-fully-initialized sessions (`getSessionInfo() == null` early return).
- Adds a periodic stats-log line + per-interval reset, matching the sibling `droppedMsgs` / `droppedLifecycleEvents` lines. The cumulative Prometheus counter stays monotonic across resets.

**Refactors bundled in** (same stats classes, no behavior change)

- Centralized the Micrometer meter print-names in the `StatsType` enum, where the rest of the meter names already live: the new `clientDisconnects`, plus the existing `droppedLifecycleEvents`, `actors.processing`, and `droppedMsgs`. `droppedMsgs` aliases `BrokerConstants.DROPPED_MSGS` because that value doubles as a historical-stats key in `common/data`, which cannot depend on `common/stats` — only the meter-side usages route through `StatsType`. **Exported metric names are unchanged.**
- Removed a redundant `@Primary` from `MsgTypeActorProcessingMetricService` and `StatsManagerImpl`. Each interface is implemented only by a real bean and a stub gated by mutually-exclusive `@ConditionalOnProperty` (feature `enabled` true vs. false/missing), so at most one bean of each type ever exists and `@Primary` has nothing to disambiguate.

**Tests**

- New `DefaultClientDisconnectStatsTest` (3): increment increases both the local count and the Micrometer counter; different reasons all map to a single untagged series; reset clears the per-interval count while the Prometheus counter stays monotonic.
- `StatsManagerImplTest`: `getClientDisconnectStats()` returns a wired instance that registers/increments the `clientDisconnects` counter.
- `DisconnectServiceImplTest`: counter assertions folded into the existing disconnect tests — an established disconnect increments with the reason passed through, cluster takeover increments, while `ON_CONNECTION_FAILURE` and the null-session early return do not (the counter fires in lockstep with the `CLIENT_DISCONNECTED` event, so the assertions live beside the lifecycle-event assertions they mirror).
- New `SessionTakenOverIntegrationTestCase` case: end-to-end proof that a real broker-generated `ON_CLUSTER_CONFLICTING_SESSIONS` takeover bumps the `clientDisconnects` meter in a live `MeterRegistry`.

**Documentation**

- No YAML/config change (this is a runtime Micrometer meter, not a config parameter). If the Prometheus metrics reference page is maintained, add `clientDisconnects_total` — a broker-wide cumulative disconnect counter.

**Backward compatibility**

- Fully backward compatible — additive new meter. The bundled refactors leave all exported metric names unchanged, and nothing is renamed or removed on the wire; no configuration or schema change.

## General checklist

- [x] Description references specific issue.

- [x] N/A — back-end only, no UI or widget/API changes.

## Back-End feature checklist

- [x] Added corresponding unit test(s): `DefaultClientDisconnectStatsTest` (new), additions to `StatsManagerImplTest` and `DisconnectServiceImplTest`, and a new end-to-end `SessionTakenOverIntegrationTestCase` case.

### DIFF
--- a/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/disconnect/DisconnectServiceImpl.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/disconnect/DisconnectServiceImpl.java
@@ -28,6 +28,7 @@ import org.thingsboard.mqtt.broker.common.data.ClientInfo;
 import org.thingsboard.mqtt.broker.common.data.SessionInfo;
 import org.thingsboard.mqtt.broker.service.auth.AuthorizationRuleService;
 import org.thingsboard.mqtt.broker.service.integration.IntegrationLifecycleEventPublisher;
+import org.thingsboard.mqtt.broker.service.stats.StatsManager;
 import org.thingsboard.mqtt.broker.service.historical.stats.TbMessageStatsReportClient;
 import org.thingsboard.mqtt.broker.service.limits.RateLimitService;
 import org.thingsboard.mqtt.broker.service.mqtt.MqttMessageGenerator;
@@ -62,6 +63,7 @@ public class DisconnectServiceImpl implements DisconnectService {
     private final TbMessageStatsReportClient tbMessageStatsReportClient;
     private final ChannelBackpressureManager channelBackpressureManager;
     private final IntegrationLifecycleEventPublisher integrationLifecycleEventPublisher;
+    private final StatsManager statsManager;
 
     @Override
     public void disconnect(ClientActorStateInfo actorState, MqttDisconnectMsg disconnectMsg) {
@@ -95,6 +97,11 @@ public class DisconnectServiceImpl implements DisconnectService {
         // CLIENT_CONNECTION_FAILED event covers that case instead.
         if (reasonType != DisconnectReasonType.ON_CONNECTION_FAILURE) {
             integrationLifecycleEventPublisher.publishDisconnected(sessionCtx, reasonType);
+            // Count every genuine disconnect of an established session. Sharing the ON_CONNECTION_FAILURE
+            // exclusion (and running past the getSessionInfo() == null early return above) keeps clientDisconnect
+            // consistent with the CLIENT_DISCONNECTED lifecycle event and avoids double-counting broker-refused
+            // connections, which are surfaced separately as connection refusals.
+            statsManager.getClientDisconnectStats().increment(reasonType);
         }
         cleanupClientSession(actorState, disconnectMsg, sessionExpiryInterval);
     }
--- a/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/disconnect/DisconnectServiceImpl.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/disconnect/DisconnectServiceImpl.java
@@ -18,6 +18,7 @@ package org.thingsboard.mqtt.broker.actors.client.service.disconnect;
 import io.netty.handler.codec.mqtt.MqttProperties;
 import io.netty.handler.codec.mqtt.MqttReasonCodes;
 import io.netty.handler.codec.mqtt.MqttVersion;
+import jakarta.annotation.PostConstruct;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
@@ -27,9 +28,8 @@ import org.thingsboard.mqtt.broker.actors.client.state.ClientActorStateInfo;
 import org.thingsboard.mqtt.broker.common.data.ClientInfo;
 import org.thingsboard.mqtt.broker.common.data.SessionInfo;
 import org.thingsboard.mqtt.broker.service.auth.AuthorizationRuleService;
-import org.thingsboard.mqtt.broker.service.integration.IntegrationLifecycleEventPublisher;
-import org.thingsboard.mqtt.broker.service.stats.StatsManager;
 import org.thingsboard.mqtt.broker.service.historical.stats.TbMessageStatsReportClient;
+import org.thingsboard.mqtt.broker.service.integration.IntegrationLifecycleEventPublisher;
 import org.thingsboard.mqtt.broker.service.limits.RateLimitService;
 import org.thingsboard.mqtt.broker.service.mqtt.MqttMessageGenerator;
 import org.thingsboard.mqtt.broker.service.mqtt.client.event.ClientSessionEventService;
@@ -38,6 +38,8 @@ import org.thingsboard.mqtt.broker.service.mqtt.flow.control.FlowControlService;
 import org.thingsboard.mqtt.broker.service.mqtt.keepalive.KeepAliveService;
 import org.thingsboard.mqtt.broker.service.mqtt.persistence.MsgPersistenceManager;
 import org.thingsboard.mqtt.broker.service.mqtt.will.LastWillService;
+import org.thingsboard.mqtt.broker.service.stats.ClientDisconnectStats;
+import org.thingsboard.mqtt.broker.service.stats.StatsManager;
 import org.thingsboard.mqtt.broker.session.ClientSessionCtx;
 import org.thingsboard.mqtt.broker.session.DisconnectReason;
 import org.thingsboard.mqtt.broker.session.DisconnectReasonType;
@@ -64,6 +66,15 @@ public class DisconnectServiceImpl implements DisconnectService {
     private final ChannelBackpressureManager channelBackpressureManager;
     private final IntegrationLifecycleEventPublisher integrationLifecycleEventPublisher;
     private final StatsManager statsManager;
+
+    private ClientDisconnectStats clientDisconnectStats;
+
+    @PostConstruct
+    void init() {
+        // Resolve once: the instance is created in StatsManagerImpl.init() and never reassigned, matching the
+        // other hot-path consumers (TbMessageStatsReportClientImpl, IntegrationLifecycleEventPublisherImpl).
+        clientDisconnectStats = statsManager.getClientDisconnectStats();
+    }
 
     @Override
     public void disconnect(ClientActorStateInfo actorState, MqttDisconnectMsg disconnectMsg) {
@@ -97,11 +108,8 @@ public class DisconnectServiceImpl implements DisconnectService {
         // CLIENT_CONNECTION_FAILED event covers that case instead.
         if (reasonType != DisconnectReasonType.ON_CONNECTION_FAILURE) {
             integrationLifecycleEventPublisher.publishDisconnected(sessionCtx, reasonType);
-            // Count every genuine disconnect of an established session. Sharing the ON_CONNECTION_FAILURE
-            // exclusion (and running past the getSessionInfo() == null early return above) keeps clientDisconnects
-            // consistent with the CLIENT_DISCONNECTED lifecycle event and avoids double-counting broker-refused
-            // connections, which are surfaced separately as connection refusals.
-            statsManager.getClientDisconnectStats().increment(reasonType);
+            // clientDisconnects counts every established-session disconnect, in lockstep with the CLIENT_DISCONNECTED event above.
+            clientDisconnectStats.increment(reasonType);
         }
         cleanupClientSession(actorState, disconnectMsg, sessionExpiryInterval);
     }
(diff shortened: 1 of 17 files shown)


common/stats/src/main/java/org/thingsboard/mqtt/broker/common/stats/StatsConstantNames.java
```diff
@@ -54,5 +54,6 @@ public class StatsConstantNames {
     public static final String MSG_TYPE = "msgType";
 
     public static final String DROPPED_LIFECYCLE_EVENTS = "droppedLifecycleEvents";
+    public static final String CLIENT_DISCONNECT = "clientDisconnect";
```
dmytro-landiak: The sibling counters in this family are plural — `droppedMsgs`, `droppedLifecycleEvents` (and `BrokerConstants.DROPPED_MSGS`) — because they count occurrences, whereas `clientDisconnect` is singular and reads more like an event name. Worth considering `clientDisconnects` for consistency, especially since this is an exported metric name (`clientDisconnect_total`) that's awkward to change once dashboards/alerts depend on it.

application/src/test/java/org/thingsboard/mqtt/broker/actors/client/service/disconnect/DisconnectServiceImplTest.java
```diff
@@ -117,6 +122,9 @@ public void setUp() {
         when(clientActorState.getQueuedMessages()).thenReturn(queuedMqttMessages);
 
         when(ctx.getClientId()).thenReturn(CLIENT_ID);
+
+        clientDisconnectStats = mock(ClientDisconnectStats.class);
```
dmytro-landiak, line 126: All the new coverage is unit-level with a mocked `StatsManager`/`ClientDisconnectStats`, so nothing asserts end to end that a real connect+disconnect actually bumps the `clientDisconnect` series. There's precedent for a live-registry assertion: `RateLimitsIntegrationTestCase` checks the sibling `droppedMsgs` via `meterRegistry.get(...).counter().count()`, and `SessionTakenOverIntegrationTestCase` already drives the `ON_CLUSTER_CONFLICTING_SESSIONS` path with real MQTT clients. Extending one of those with a counter assertion would catch wiring regressions the mocks can't. Not blocking, but it looks achievable given the existing suites.

application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/disconnect/DisconnectServiceImpl.java
```diff
@@ -95,6 +97,11 @@ public void disconnect(ClientActorStateInfo actorState, MqttDisconnectMsg discon
         // CLIENT_CONNECTION_FAILED event covers that case instead.
         if (reasonType != DisconnectReasonType.ON_CONNECTION_FAILURE) {
             integrationLifecycleEventPublisher.publishDisconnected(sessionCtx, reasonType);
+            // Count every genuine disconnect of an established session. Sharing the ON_CONNECTION_FAILURE
+            // exclusion (and running past the getSessionInfo() == null early return above) keeps clientDisconnect
+            // consistent with the CLIENT_DISCONNECTED lifecycle event and avoids double-counting broker-refused
+            // connections, which are surfaced separately as connection refusals.
+            statsManager.getClientDisconnectStats().increment(reasonType);
```
dmytro-landiak: The other hot-path consumers of a `StatsManager`-vended stats object cache it once instead of resolving it per call — `TbMessageStatsReportClientImpl` and `IntegrationLifecycleEventPublisherImpl` grab `droppedMsgStats` / `droppedLifecycleEventStats` at init and then just call `.increment()`. Here `getClientDisconnectStats()` is resolved inline on every disconnect. The cost is negligible, but since the instance is created once in `StatsManagerImpl.init()` and never reassigned, caching the reference would line up with the convention this otherwise mirrors closely. Your call — inline is fine functionally.

application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/disconnect/DisconnectServiceImpl.java
```diff
@@ -95,6 +97,11 @@ public void disconnect(ClientActorStateInfo actorState, MqttDisconnectMsg discon
         // CLIENT_CONNECTION_FAILED event covers that case instead.
         if (reasonType != DisconnectReasonType.ON_CONNECTION_FAILURE) {
             integrationLifecycleEventPublisher.publishDisconnected(sessionCtx, reasonType);
+            // Count every genuine disconnect of an established session. Sharing the ON_CONNECTION_FAILURE
```
dmytro-landiak: This 4-line comment largely re-derives the `ON_CONNECTION_FAILURE` exclusion rationale that the 6-line comment directly above the `if` already spells out. Two back-to-back explanations of overlapping reasoning is a lot of prose for a one-line increment — could this be trimmed to something like `// clientDisconnect mirrors the CLIENT_DISCONNECTED event above`, leaning on the existing comment for the why?

application/src/test/java/org/thingsboard/mqtt/broker/actors/client/service/disconnect/DisconnectServiceImplTest.java
```diff
@@ -240,6 +248,43 @@ public void cleanupClientSession_callsReleasePublishedInFlightCtx_beforeRemoveFr
...
+    @Test
+    public void givenEstablishedSession_whenDisconnect_thenClientDisconnectCounterIncrementedWithReason() {
+        MqttDisconnectMsg disconnectMsg = newDisconnectMsg(new DisconnectReason(ON_DISCONNECT_MSG));
+        disconnectService.disconnect(clientActorState, disconnectMsg);
+
+        verify(clientDisconnectStats, times(1)).increment(ON_DISCONNECT_MSG);
+    }
+
+    @Test
+    public void givenClusterConflictingSession_whenDisconnect_thenClientDisconnectCounterIncremented() {
```
dmytro-landiak: This test and the two below it (`givenConnectionFailure_...`, `givenSessionInfoIsNull_...`) reuse the exact arrange/act of existing tests in this file (`givenClusterConflictingSession_whenDisconnect_thenEmitLifecycleDisconnectButSkipSessionEvent`, `givenConnectionFailure_whenDisconnect_thenDoesNotEmitLifecycleDisconnect`, `givenSessionInfoIsNull_whenDisconnectClient_thenCloseChannel`) and only add one more `verify`. Since the counter shares the same three code paths as the lifecycle publish it mirrors, folding the `verify` into those existing tests would avoid the duplicated message construction. Counter-argument: separate focused tests document the metric behavior on its own — reasonable either way, your call.

application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/disconnect/DisconnectServiceImpl.java
```diff
@@ -28,6 +28,7 @@
 import org.thingsboard.mqtt.broker.common.data.SessionInfo;
 import org.thingsboard.mqtt.broker.service.auth.AuthorizationRuleService;
 import org.thingsboard.mqtt.broker.service.integration.IntegrationLifecycleEventPublisher;
+import org.thingsboard.mqtt.broker.service.stats.StatsManager;
```
dmytro-landiak: This `StatsManager` import is slightly out of order — it landed between `service.integration` and `service.historical.stats`, whereas `service.stats` should sort after the `service.mqtt.*` block. Every other file that imports `StatsManager` (including this PR's own test file) places it there; an organize-imports pass would line it up.

dmytro-landiak (commented):
## Review summary

Reviewed 11 changed files in [feat(stats): add clientDisconnect Prometheus counter](https://github.com/thingsboard/tbmq/pull/344). The change is correct and safe: the new `ClientDisconnectStats` interface is implemented by both `StatsManager` impls (real + stub, so the getter never returns null), the increment is correctly scoped inside the `!= ON_CONNECTION_FAILURE` block and after the `getSessionInfo() == null` early return — so it fires exactly when the `CLIENT_DISCONNECTED` lifecycle event does — and the counter is thread-safe. No correctness or safety issues found.

Left 6 comments inline, all quality/consistency nits. The two most worth a look before merge: the singular `clientDisconnect` metric name vs. the plural sibling counters (`droppedMsgs`, `droppedLifecycleEvents`) — awkward to change once it's an exported series — and the absence of an end-to-end assertion that a real disconnect bumps the counter (there's existing integration-test precedent for the sibling `droppedMsgs`).

---

application/src/main/java/org/thingsboard/mqtt/broker/service/stats/ClientDisconnectStats.java
```diff
@@ -0,0 +1,46 @@
...
+     */
+    void increment(DisconnectReasonType reasonType);
+
+    /**
+     * Number of disconnects since the last {@link #reset()} — read for the periodic stats log line.
+     */
+    int getCount();
+
+    /**
+     * Clears the per-interval count read by {@link #getCount()}. The cumulative {@code clientDisconnect}
```
dmytro-landiak: The rename to `clientDisconnects` updated the class-level javadoc but missed this `reset()` line — it still reads `{@code clientDisconnect}` (the old singular name). Worth updating so the shipped interface doc is internally consistent.

common/stats/src/main/java/org/thingsboard/mqtt/broker/common/stats/StatsType.java
```diff
@@ -68,6 +70,10 @@ public enum StatsType {
     INTEGRATION("integration"),
     INTEGRATION_PROCESSOR("integrationProcessor"),
     INTEGRATION_EVENT_PROCESSOR("integrationEventProcessor"),
+
+    DROPPED_MSGS(BrokerConstants.DROPPED_MSGS),
```
dmytro-landiak, line 74: Aliasing the print name to `BrokerConstants.DROPPED_MSGS` here to keep it equal to the historical key reads well. The pre-existing `SUBSCRIPTIONS("subscriptions")` entry a few lines up documents the same tie to `BrokerConstants.SUBSCRIPTIONS` in a comment but hardcodes the literal instead. Since this pass is centralizing print-names, it might be worth having `SUBSCRIPTIONS` reference the constant too, so the same invariant is expressed one way rather than two.

dmytro-landiak (commented):
## Re-review summary

Re-reviewed [feat(stats): add clientDisconnects Prometheus counter](https://github.com/thingsboard/tbmq/pull/344) — verified 6 finding(s) from [previous review](https://github.com/thingsboard/tbmq/pull/344#pullrequestreview-4651756522).

| Status | Count |
|--------|-------|
| ✅ Resolved | 6 |
| 💬 Acknowledged | 0 |
| ❌ Unresolved | 0 |

All previous findings have been addressed. Also found 2 new minor issues in the fix commits, commented inline.

### Finding details

- ✅ **StatsConstantNames.java** — singular `clientDisconnect` vs. the plural sibling counters — *Fixed: renamed to `clientDisconnects` and moved the print-name into the `StatsType` enum.*
- ✅ **DisconnectServiceImplTest.java** — no end-to-end assertion that a real disconnect bumps the series — *Fixed: added a case in `SessionTakenOverIntegrationTestCase` asserting a real takeover disconnect increments the live `clientDisconnects` meter.*
- ✅ **DisconnectServiceImpl.java** — stats object resolved inline on every disconnect — *Fixed: cached once in `@PostConstruct`, matching the sibling hot-path consumers.*
- ✅ **DisconnectServiceImpl.java** — redundant 4-line increment-site comment — *Fixed: trimmed to a one-liner leaning on the block above.*
- ✅ **DisconnectServiceImplTest.java** — duplicated counter unit tests — *Fixed: folded the assertions into the existing disconnect tests and removed the clones.*
- ✅ **DisconnectServiceImpl.java** — `StatsManager` import out of order — *Fixed: import block re-sorted.*

---

PR #345, opened 2026-07-08
refactor(stats): unify actor-processing metrics for client and device actors

dmytro-landiak:
## Pull Request description

Unifies TBMQ's actor message-processing metrics behind a single, `StatsType`-parameterized `ActorStats`, and extends full metric coverage (processing time **and** mailbox queue-wait time) to the persisted device actor.

Previously there were two overlapping mechanisms:
- a generic `MsgTypeActorProcessingMetricService` — metric `actors.processing`, Prometheus-only, gated by a default-off flag, with a `StopWatch` that ran on **every** actor message even when the feature was disabled; and
- a client-only `ClientActorStats` — metrics `clientActor.processing.time` / `clientActor.msgInQueueTime`, reported to both the periodic stats logs and Prometheus.

The generic mechanism was redundant for `ClientActor` and was the *only* (default-off) coverage for `PersistedDeviceActor`, which otherwise had no processing metrics at all.

### Scope of changes
- Replace the client-only `ClientActorStats` / `DefaultClientActorStats` / `StubClientActorStats` with a parameterized `ActorStats` / `DefaultActorStats(StatsFactory, StatsType)` / `StubActorStats` (one instance per actor type), following the existing `IntegrationProcessorStatsImpl(StatsFactory, StatsType)` pattern.
- Centralize timing in `ContextAwareActor.process()`: processing time is always recorded; queue-wait time is recorded when the message carries a creation timestamp (`TimedMsg`). This removes the per-message `StopWatch` that previously ran unconditionally on the hottest path.
- Make all persisted-device actor messages `TimedMsg` (promote `AbstractTimedMsg` to `actors/shared`), so `PersistedDeviceActor` now reports queue-wait time as well.
- Register and periodically log both a client and a device `ActorStats` via `StatsManager`.
- Remove the redundant `MsgTypeActorProcessingMetricService`, its `ActorProcessingMetricService` interface and stub, and the `actors.system.processing-metrics` config flag.

### Metrics impact
- **Unchanged:** `clientActor.processing.time{msgType}`, `clientActor.msgInQueueTime`.
- **New:** `deviceActor.processing.time{msgType}`, `deviceActor.msgInQueueTime`.
- **Removed (breaking, but default-off):** the `actors.processing{msgType}` Prometheus series and the `ACTORS_SYSTEM_PROCESSING_METRICS_ENABLED` environment variable.

All actor metrics remain governed by the global `stats.enabled` switch and are reported to both the periodic stats logs and Prometheus.

### Backward compatibility
No database schema, queue, or proto changes — rolling-upgrade safe. The only externally visible change is the removal of the default-off `actors.processing` metric series and its environment variable, noted above.

### Documentation notes
- Remove `ACTORS_SYSTEM_PROCESSING_METRICS_ENABLED` from the configuration docs.
- Add the new `deviceActor.processing.time` / `deviceActor.msgInQueueTime` metrics alongside the existing `clientActor.*` entries.

## General checklist

- [x] Changes are backward compatible or upgrade script is provided. (No schema/queue/proto change; the only breaking item — removal of the default-off `actors.processing` series and its env var — is documented above.)

- [x] N/A — backend-only change; no front-end code affected, so no screenshots or UI API changes.

## Back-End feature checklist

- [x] Added corresponding unit test(s): `DefaultActorStatsTest` (new), `ContextAwareActorTest` (rewritten to cover the centralized timing dispatch), and `StatsManagerImplTest` (existing coverage preserved).


application/src/main/java/org/thingsboard/mqtt/broker/actors/device/messages/ChannelWritableEventMsg.java
```diff
@@ -20,11 +20,12 @@
 import lombok.extern.slf4j.Slf4j;
 import org.thingsboard.mqtt.broker.actors.msg.MsgType;
 import org.thingsboard.mqtt.broker.actors.msg.TbActorMsg;
+import org.thingsboard.mqtt.broker.actors.shared.AbstractTimedMsg;
 
 @Slf4j
 @Getter
 @RequiredArgsConstructor
-public class ChannelWritableEventMsg implements TbActorMsg {
+public class ChannelWritableEventMsg extends AbstractTimedMsg implements TbActorMsg {
```
dmytro-landiak, line 28: `AbstractTimedMsg` captures `createdTimeNanos` once in its field initializer, so it implicitly assumes a fresh instance per send. But three of the device messages promoted to `TimedMsg` here are reused singletons — `ChannelWritableEventMsg.INSTANCE`, `ChannelNonWritableEventMsg.INSTANCE`, and `RemovePersistedMessagesEventMsg.DEFAULT` (dispatched from `DeviceActorManagerImpl` / `PersistedDeviceActorMessageProcessor`). For those, `getMsgCreatedTimeNanos()` returns the class-load time, so `ContextAwareActor.process()` records `now - classLoadTime` into the new `deviceActor.msgInQueueTime` timer every time one is processed — i.e. its max/avg trend toward JVM uptime rather than real mailbox latency (`ResettableTimer` feeds both the Prometheus timer and the periodic stats log). The two client channel singletons (`WritableChannelMsg.DEFAULT` / `NonWritableChannelMsg.DEFAULT`) already have this latent behavior, so it's at least consistent — but since this PR is specifically about metric correctness, it may be worth either not making these control singletons `TimedMsg`, or documenting the fresh-instance-per-send contract on `AbstractTimedMsg` so the next singleton doesn't silently skew the metric.

application/src/main/java/org/thingsboard/mqtt/broker/actors/service/ContextAwareActor.java
```diff
@@ -16,38 +16,40 @@
...
+
+import java.util.concurrent.TimeUnit;
 
 @Slf4j
 public abstract class ContextAwareActor extends AbstractTbActor {
 
     protected final ActorSystemContext systemContext;
-    private final StopWatch stopWatch;
-    private final ActorProcessingMetricService actorProcessingMetricService;
+    protected final ActorStats actorStats;
```
dmytro-landiak: `actorStats` is `protected`, but neither subclass in this module (`ClientActor`, `PersistedDeviceActor`) references it anymore now that all timing lives in the base `process()`. Could this be `private`? If a divergent PE actor subclass reads it, `protected` is justified — worth a quick check before tightening.

application/src/test/java/org/thingsboard/mqtt/broker/service/stats/DefaultActorStatsTest.java
```diff
@@ -0,0 +1,108 @@
...
+        stats.logMsgQueueTime(new TestTimedMsg(System.nanoTime() - 2_000_000L), TimeUnit.NANOSECONDS);
+
+        Timer timer = meterRegistry.find("deviceActor.msgInQueueTime").timer();
+        assertNotNull(timer);
+        assertEquals(1, timer.count());
+        assertEquals(1, stats.getMsgCount());
+    }
+
+    @Test
+    public void deviceMessagesAreTimedMsg() {
```
dmytro-landiak: This guards a nice invariant — "device messages are `TimedMsg`, so queue time gets recorded" — but it only spot-checks 3 of the 12 device message classes. Someone adding a new device message who forgets to extend `AbstractTimedMsg` would get no queue-time recording and no failing test. Could this enumerate all 12 (or scan the `device.messages` package) so the invariant is enforced rather than sampled?

application/src/test/java/org/thingsboard/mqtt/broker/service/stats/DefaultActorStatsTest.java
```diff
@@ -0,0 +1,108 @@
...
+    @Before
+    public void setUp() {
+        meterRegistry = new SimpleMeterRegistry();
+        statsFactory = new DefaultStatsFactory(meterRegistry);
+    }
+
+    @Test
+    public void givenClientActorType_whenLogProcessingTime_thenTimerNamedClientActor() {
+        ActorStats stats = new DefaultActorStats(statsFactory, StatsType.CLIENT_ACTOR);
+        stats.logMsgProcessingTime(MsgType.INCOMING_PUBLISH_MSG, System.nanoTime() - 1_500_000L, TimeUnit.NANOSECONDS);
```
dmytro-landiak: This test and `givenDeviceActorType_whenLogProcessingTime_thenTimerNamedDeviceActor` are near-identical, differing only in the `StatsType` and the expected timer name — a good candidate for a single parameterized test. The copy has also drifted: this one passes `System.nanoTime() - 1_500_000L` as the start time while the device variant passes `System.nanoTime()`, yet neither asserts on the recorded duration (both only check `count == 1`), so the 1.5ms offset reads as accidental noise.

dmytro-landiak (commented):
## Review summary

Reviewed 42 changed files in [refactor(stats): unify actor-processing metrics for client and device actors](https://github.com/thingsboard/tbmq/pull/345). Left 4 comments inline.

The refactor is clean and well-scoped: the `StatsType`-parameterized `ActorStats` mirrors the existing `IntegrationProcessorStatsImpl` pattern, timing is correctly centralized in `ContextAwareActor.process()` with no double-counting (`ClientActor`'s old inline timing was fully removed), dropping the always-on `StopWatch` is a genuine hot-path win, and a straggler check found no dangling references to the removed `ActorProcessingMetricService` / `actors.processing` metric / config flag. The main thing worth a look is that a few reused singleton device messages now carry a static-init creation timestamp, which will skew the new `deviceActor.msgInQueueTime` metric.

## Additional findings

- **application/src/main/java/org/thingsboard/mqtt/broker/service/stats/ActorStats.java:29** — Now that this interface is renamed and consumed by both actor types, it may be worth tightening two pre-existing API shapes. The `TimeUnit unit` parameter on `logMsgProcessingTime`/`logMsgQueueTime` is effectively decorative: `DefaultActorStats` always computes the delta from `System.nanoTime()` (nanoseconds) and only then applies `unit`, so passing anything other than `NANOSECONDS` silently mis-scales the recorded value — and every caller already hard-codes `NANOSECONDS`. And `logMsgQueueTime(TbActorMsg)` casts its argument to `TimedMsg` internally, while its only caller (`ContextAwareActor`) already guards with `instanceof TimedMsg`; accepting `TimedMsg` directly would make the contract type-safe rather than a runtime cast.

---

application/src/test/java/org/thingsboard/mqtt/broker/service/stats/DefaultActorStatsTest.java
```diff
@@ -0,0 +1,114 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.thingsboard.mqtt.broker.service.stats;
+
+import io.micrometer.core.instrument.Timer;
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+import org.junit.jupiter.api.BeforeEach;
```
dmytro-landiak, line 20: Migrating this file to JUnit 5 leaves it as the only Jupiter test in the `service/stats` package (the other six are JUnit 4), and its same-PR sibling `ContextAwareActorTest` — also touched by these fix commits — stays on JUnit 4, so the PR ships two adjacent actor-stats tests on different engines. Jupiter has plenty of precedent elsewhere in the module, so this isn't wrong — just a consistency wrinkle. The `@ParameterizedTest` collapse is the real win here and is the natural reason to move to Jupiter; if standardizing was the intent, aligning `ContextAwareActorTest` too would avoid the split within one PR — otherwise keeping this file on JUnit 4 like its package neighbors would read more consistently.

dmytro-landiak (commented):
## Re-review summary

Re-reviewed [refactor(stats): unify actor-processing metrics for client and device actors](https://github.com/thingsboard/tbmq/pull/345) — verified 5 findings from the [previous review](https://github.com/thingsboard/tbmq/pull/345#pullrequestreview-4653331291).

| Status | Count |
|--------|-------|
| ✅ Resolved | 5 |
| 💬 Acknowledged | 0 |

All previous findings were addressed in code (no developer replies needed) — and in several cases beyond what was suggested. Also found 1 minor new issue in the fix commits, commented inline.

### Finding details

- ✅ **ChannelWritableEventMsg.java** — reused singleton's static-init `createdTimeNanos` skewed `deviceActor.msgInQueueTime` — *Fixed: removed all singleton instances (device **and** the client ones), construct a fresh instance per send, and documented the "fresh instance per send" contract on `AbstractTimedMsg`.*
- ✅ **ContextAwareActor.java** — `actorStats` was `protected` but unused by subclasses — *Fixed: now `private`.*
- ✅ **DefaultActorStatsTest.java** — the `TimedMsg` invariant test spot-checked only 3 of 12 device messages — *Fixed: now classpath-scans the whole `device.messages` package, with a non-empty guard.*
- ✅ **DefaultActorStatsTest.java** — two near-duplicate tests with a drifted start-time argument — *Fixed: collapsed into one `@ParameterizedTest`; the drift is gone.*
- ✅ **ActorStats.java** — decorative `TimeUnit` parameter and an internal `TimedMsg` cast — *Fixed: dropped the parameter and `logMsgQueueTime` now takes `TimedMsg` directly.*

---

PR #346, opened 2026-07-09
feat(stats): complete droppedMsgs across the PUBLISH pipeline

dmytro-landiak:
## Pull Request description

Makes the broker-wide `droppedMsgs` counter count every PUBLISH the broker **accepted but permanently failed to deliver — exactly once** — and stop counting recoverable/persistent losses.

**Scope of changes** — counting added/corrected at each PUBLISH-pipeline site:

| Site | Change |
|---|---|
| `MqttPublishHandler` | receive-maximum-exceeded QoS1/2 drops |
| `PublishMsgConsumerServiceImpl` | inbound give-up commit: leftover pending + failed |
| `DefaultMqttPublishMsgDeliveryService` | sync write-throw **and** async write-future failures, via `isCountableDrop` |
| `PublishedInFlightCtxImpl` | in-flight/delay-queue overflow + delayed-TTL drops (flow-control `dropsOverflow`/`dropsTtl` meters unchanged) |
| `ApplicationPersistenceProcessorImpl` | APPLICATION give-up: un-acked PUBLISH (`publishPendingMap`, excludes PUBREL) |
| `DeviceMsgQueueConsumerImpl` | DEVICE give-up: failed-persist messages (`failedMap`) |

**Counting principle** (`isCountableDrop`): count a delivery / flow-control drop only if **non-retained** and (**non-persistent session** or **QoS0**). A persistent-session QoS>0 copy is recoverable from the store (APPLICATION → Kafka, counted once at the APPLICATION give-up; DEVICE → redelivered from Redis); QoS0 is never stored; retained is recoverable from the retained store.

**Deliberately unchanged:** `DuplexTrafficHandler` too-long-frame, `MsgDispatcherServiceImpl` zero-subscriber, `BasicDownLinkProcessorImpl` session-gone, MQTT5 message-expiry.

**Documentation:** no user-facing documentation change required — `droppedMsgs` is a pre-existing untagged metric and only its coverage is broadened. Deferred follow-ups (out of scope, tracked separately): a distinct `expiredMsgs` meter for MQTT5-expiry discards.

## General checklist

  - _N/A — back-end only, no UI change._
  - _N/A — no widget / API change._

## Back-End feature checklist

  - _Unit tests (JUnit 4/5 + Mockito) added/updated for all six sites — 111 tests across the 7 touched classes pass; each drop path asserted via `verify(...)` on `TbMessageStatsReportClient`. No integration test: the counting is deterministic and fully covered at unit level._
  - _N/A — no new dependency added._

### DIFF
--- a/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/state/PublishedInFlightCtxImpl.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/state/PublishedInFlightCtxImpl.java
@@ -275,7 +275,11 @@ public class PublishedInFlightCtxImpl implements PublishedInFlightCtx {
     // A flow-control drop counts toward droppedMsgs only for a non-persistent session and a non-retained
     // message: persistent copies are recoverable from the store (APPLICATION via Kafka, counted once at
     // give-up in ApplicationPersistenceProcessorImpl; DEVICE via Redis redelivery), and retained messages are
-    // recoverable from the retained-message store (consistent with DefaultMqttPublishMsgDeliveryService).
+    // recoverable from the retained-message store.
+    //
+    // NOTE: this predicate intentionally has NO QoS0 clause, unlike DefaultMqttPublishMsgDeliveryService#isCountableDrop.
+    // QoS0 messages short-circuit in atMostOnce()/addInFlightMsg() and never enter the delay queue, so they cannot be
+    // dropped here — the two predicates are deliberately different. Do not "align" them by adding a QoS0 clause here.
     private boolean isCountableDrop(MqttPublishMessage msg) {
         return !clientSessionCtx.getSessionInfo().isPersistent() && !msg.fixedHeader().isRetain();
     }
--- a/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/state/PublishedInFlightCtxImpl.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/state/PublishedInFlightCtxImpl.java
@@ -218,14 +218,22 @@ public class PublishedInFlightCtxImpl implements PublishedInFlightCtx {
             lock.unlock();
         }
 
+        // Session persistence is fixed for the session, so evaluate it once here rather than per expired message:
+        // for a persistent session nothing expiring here is a countable drop (the copy is recoverable from the
+        // store), so the per-message retain check is skipped entirely. Count in the same pass that releases the
+        // buffers instead of a second stream over the list. See isCountableDrop for the full rule and rationale.
+        boolean countableSession = !toRelease.isEmpty() && !clientSessionCtx.getSessionInfo().isPersistent();
+        int countableDrops = 0;
         for (MqttPublishMessage m : toRelease) {
+            if (countableSession && !m.fixedHeader().isRetain()) {
+                countableDrops++;
+            }
             ReferenceCountUtil.safeRelease(m);
         }
         int expiredCount = toRelease.size();
         if (expiredCount > 0) {
             stats.decDelayed(expiredCount);
             stats.incDropTtl(expiredCount);
-            int countableDrops = (int) toRelease.stream().filter(this::isCountableDrop).count();
             if (countableDrops > 0) {
                 tbMessageStatsReportClient.reportDroppedMsgs(countableDrops);
             }
--- a/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/state/PublishedInFlightCtxImpl.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/state/PublishedInFlightCtxImpl.java
@@ -118,6 +118,8 @@ public class PublishedInFlightCtxImpl implements PublishedInFlightCtx {
                     clientId, clientReceiveMax, delayedMsgQueueMaxSize);
             ReferenceCountUtil.safeRelease(toRelease);
             stats.incDropOverflow();
+            // Count as dropped only for non-persistent sessions; persistent (APPLICATION) copies stay in the
+            // store and are counted once at give-up (ApplicationPersistenceProcessorImpl).
             if (!clientSessionCtx.getSessionInfo().isPersistent()) {
                 tbMessageStatsReportClient.reportDroppedMsgs();
             }
@@ -225,6 +227,8 @@ public class PublishedInFlightCtxImpl implements PublishedInFlightCtx {
         if (expiredCount > 0) {
             stats.decDelayed(expiredCount);
             stats.incDropTtl(expiredCount);
+            // Count as dropped only for non-persistent sessions; persistent (APPLICATION) copies stay in the
+            // store and are counted once at give-up (ApplicationPersistenceProcessorImpl).
             if (!clientSessionCtx.getSessionInfo().isPersistent()) {
                 tbMessageStatsReportClient.reportDroppedMsgs(expiredCount);
             }
--- a/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/state/PublishedInFlightCtxImpl.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/state/PublishedInFlightCtxImpl.java
@@ -118,9 +118,7 @@ public class PublishedInFlightCtxImpl implements PublishedInFlightCtx {
                     clientId, clientReceiveMax, delayedMsgQueueMaxSize);
             ReferenceCountUtil.safeRelease(toRelease);
             stats.incDropOverflow();
-            // Count as dropped only for non-persistent sessions; persistent (APPLICATION) copies stay in the
-            // store and are counted once at give-up (ApplicationPersistenceProcessorImpl).
-            if (!clientSessionCtx.getSessionInfo().isPersistent()) {
+            if (isCountableDrop(toRelease)) {
                 tbMessageStatsReportClient.reportDroppedMsgs();
             }
         }
@@ -227,10 +225,9 @@ public class PublishedInFlightCtxImpl implements PublishedInFlightCtx {
         if (expiredCount > 0) {
             stats.decDelayed(expiredCount);
             stats.incDropTtl(expiredCount);
-            // Count as dropped only for non-persistent sessions; persistent (APPLICATION) copies stay in the
-            // store and are counted once at give-up (ApplicationPersistenceProcessorImpl).
-            if (!clientSessionCtx.getSessionInfo().isPersistent()) {
-                tbMessageStatsReportClient.reportDroppedMsgs(expiredCount);
+            int countableDrops = (int) toRelease.stream().filter(this::isCountableDrop).count();
+            if (countableDrops > 0) {
+                tbMessageStatsReportClient.reportDroppedMsgs(countableDrops);
             }
         }
 
@@ -273,6 +270,14 @@ public class PublishedInFlightCtxImpl implements PublishedInFlightCtx {
 
     private boolean atMostOnce(MqttPublishMessage mqttPubMsg) {
         return MqttQoS.AT_MOST_ONCE == mqttPubMsg.fixedHeader().qosLevel();
+    }
+
+    // A flow-control drop counts toward droppedMsgs only for a non-persistent session and a non-retained
+    // message: persistent copies are recoverable from the store (APPLICATION via Kafka, counted once at
+    // give-up in ApplicationPersistenceProcessorImpl; DEVICE via Redis redelivery), and retained messages are
+    // recoverable from the retained-message store (consistent with DefaultMqttPublishMsgDeliveryService).
+    private boolean isCountableDrop(MqttPublishMessage msg) {
+        return !clientSessionCtx.getSessionInfo().isPersistent() && !msg.fixedHeader().isRetain();
     }
 
 }
--- a/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/state/PublishedInFlightCtxImpl.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/actors/client/state/PublishedInFlightCtxImpl.java
@@ -20,6 +20,7 @@ import io.netty.handler.codec.mqtt.MqttQoS;
 import io.netty.util.ReferenceCountUtil;
 import lombok.extern.slf4j.Slf4j;
 import org.thingsboard.mqtt.broker.common.data.mqtt.MqttPubMsgWithCreatedTime;
+import org.thingsboard.mqtt.broker.service.historical.stats.TbMessageStatsReportClient;
 import org.thingsboard.mqtt.broker.service.mqtt.delivery.MqttPublishMsgDeliveryService;
 import org.thingsboard.mqtt.broker.service.mqtt.flow.control.FlowControlService;
 import org.thingsboard.mqtt.broker.service.stats.FlowControlStats;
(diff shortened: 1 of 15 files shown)


dmytro-landiak (commented):
## Review summary

Reviewed 15 changed files in [feat(stats): complete droppedMsgs across the PUBLISH pipeline ](https://github.com/thingsboard/tbmq/pull/346). Left 4 comments inline.

The core counting logic is careful and correct — I traced every site and found no double-counting or missed-count bugs. `onFailure` moves entries between disjoint pending/failed maps, each retry round rebuilds a fresh processing context (so `failedMap` never accumulates), delivery-layer and give-up counting are mutually exclusive through the persistence/QoS `isCountableDrop` gating, and counting fires only on genuine give-up commits (verified against both `SkipStrategy` and `RetryStrategy`). I also confirmed the `statsEnabled` gate on the async listener loses no counts, since `droppedMsgStats` is a no-op stub under the same master switch. The comments below are maintainability, test-coverage, and API-ergonomics observations — not correctness defects.

---

application/src/main/java/org/thingsboard/mqtt/broker/actors/client/state/PublishedInFlightCtxImpl.java
```diff
@@ -261,4 +272,12 @@ private boolean atMostOnce(MqttPublishMessage mqttPubMsg) {
         return MqttQoS.AT_MOST_ONCE == mqttPubMsg.fixedHeader().qosLevel();
     }
 
+    // A flow-control drop counts toward droppedMsgs only for a non-persistent session and a non-retained
+    // message: persistent copies are recoverable from the store (APPLICATION via Kafka, counted once at
+    // give-up in ApplicationPersistenceProcessorImpl; DEVICE via Redis redelivery), and retained messages are
+    // recoverable from the retained-message store (consistent with DefaultMqttPublishMsgDeliveryService).
+    private boolean isCountableDrop(MqttPublishMessage msg) {
```
dmytro-landiak, line 285: There are now two `isCountableDrop` methods — this one and the one in `DefaultMqttPublishMsgDeliveryService` — that share a name but differ: this variant omits the QoS0 clause the delivery-service variant has. The divergence is actually correct (QoS0 short-circuits in `atMostOnce`/`addInFlightMsg`, so it never reaches here), but the comment just above says the rule is *"consistent with DefaultMqttPublishMsgDeliveryService"*, which a maintainer diffing the two predicates will find isn't literally true. Two identically-named predicates with subtly different logic is an easy trap — someone "aligning" them could reintroduce QoS0 counting. Might be worth either rewording the comment to state they intentionally differ (and why), or extracting one helper that takes explicit persistent/retained/qos flags.

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/delivery/DefaultMqttPublishMsgDeliveryService.java
```diff
@@ -59,10 +60,10 @@ public void sendAlreadyTrackedPublishMsgToClient(ClientSessionCtx ctx, MqttPubli
         try {
             long startTime = System.nanoTime();
             ChannelFuture future = ctx.getChannel().writeAndFlush(msg);
-            recordDeliveryOnSuccess(future, startTime);
+            recordDeliveryOutcome(ctx, msg, future, startTime);
         } catch (Exception e) {
             log.warn("[{}][{}] Failed to send PUBLISH msg to MQTT client", ctx.getClientId(), ctx.getSessionId(), e);
-            if (!msg.fixedHeader().isRetain()) {
+            if (isCountableDrop(ctx, msg)) {
```
dmytro-landiak, line 66: This synchronous-throw branch in `sendAlreadyTrackedPublishMsgToClient` picked up the new `isCountableDrop` gating, but its failure path looks untested — the suite covers the sync-throw and async-failure drops for `sendPublishMsgToClient`, yet only the success path for this method. Since this is exactly the method the flow-control drain path calls, a regression in its countable-drop decision would slip through CI. Worth mirroring one of the `sendPublishMsgToClient` drop cases here?

application/src/main/java/org/thingsboard/mqtt/broker/session/ClientSessionCtx.java
```diff
@@ -170,10 +171,11 @@ public void clearConnectMsg() {
     public void initPublishedInFlightCtx(FlowControlService flowControlService,
                                          MqttPublishMsgDeliveryService deliveryService,
                                          FlowControlStats stats,
+                                         TbMessageStatsReportClient tbMessageStatsReportClient,
```
dmytro-landiak, line 174: This new collaborator was inserted mid-list here and in the `PublishedInFlightCtxImpl` constructor, which now takes 7 positional args — two stats-ish collaborators (`FlowControlStats`, then this) followed by the adjacent `receiveMaxValue`/`delayedQueueMaxSize` ints. The object insertion is compiler-checked and there's a single caller, so misuse risk is low today, but that trailing int pair is an easy swap to make silently. Not blocking — but if this constructor gains another param, a small param object or builder would make it much harder to mis-order.

application/src/test/java/org/thingsboard/mqtt/broker/service/mqtt/persistence/application/ApplicationPersistenceProcessorImplTest.java
```diff
@@ -556,6 +583,28 @@ private void invokeProcessMainPack(ApplicationPubRelMsgCtx pubRelMsgCtx,
         }
     }
 
+    private void invokeReportSkippedMessagesIfAny(String clientId, ApplicationPackProcessingResult result) {
+        try {
+            var method = ApplicationPersistenceProcessorImpl.class.getDeclaredMethod(
+                    "reportSkippedMessagesIfAny", String.class, ApplicationPackProcessingResult.class);
```
dmytro-landiak, line 626: The new PUBLISH-drop counting is only exercised by reflecting straight into the private `reportSkippedMessagesIfAny`, which skips the `tryCommitPack` → `decision.isCommit()` gating and never proves the report fires on the real give-up path. This class already drives give-up end to end (`processMainPack_retryAll_whenClientNeverAcks_...`), but only with a never-acked PUBREL, so `skippedPublish` is always 0 there and the new behavior isn't touched. Could you reuse that harness with a never-acked PUBLISH and assert `reportDroppedMsgs(n)`, instead of the reflection shortcut? That would cover the actual commit/give-up flow.

application/src/main/java/org/thingsboard/mqtt/broker/actors/client/state/PublishedInFlightCtxImpl.java
dmytro-landiak: Reworded in a1f8dc9ea. The comment no longer claims consistency — it now states this predicate intentionally omits the QoS0 clause that `DefaultMqttPublishMsgDeliveryService#isCountableDrop` carries (QoS0 short-circuits in `atMostOnce()`/`addInFlightMsg()` and never reaches the delay queue, so it cannot be dropped here) and explicitly warns against "aligning" the two. Kept them as two separate predicates rather than extracting a shared flags helper: the call sites live in different packages and legitimately need different logic.

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/delivery/DefaultMqttPublishMsgDeliveryService.java
dmytro-landiak: Added in 7774bda90. Two sync-throw tests now cover `sendAlreadyTrackedPublishMsgToClient` (the drain path): `givenSyncThrowNonPersistentNonRetained_whenSendAlreadyTracked_thenReportsDroppedAndRecordsNothing` (countable → reported) and `givenSyncThrowPersistentQos1_whenSendAlreadyTracked_thenNotCounted` (persistent QoS>0 → not reported), mirroring the `sendPublishMsgToClient` drop cases and proving the gate in both directions.

application/src/main/java/org/thingsboard/mqtt/broker/session/ClientSessionCtx.java
dmytro-landiak: Leaving as-is for this PR. There's a single construction site, the object args are compiler-checked, and the only genuinely swap-prone part — the adjacent `receiveMaxValue`/`delayedQueueMaxSize` int pair — predates this change (this PR only inserted a distinctly-typed object param between the existing object args). A builder/param-object is out of scope for a metrics-completeness PR; better as a standalone refactor if the constructor gains another param, as flagged here.

application/src/test/java/org/thingsboard/mqtt/broker/service/mqtt/persistence/application/ApplicationPersistenceProcessorImplTest.java
dmytro-landiak: Added in 7774bda90: `processMainPack_retryAll_whenClientNeverAcksPublish_reportsDroppedOnGiveUpCommit` drives the real give-up path (RETRY_ALL with the production `RetryStrategy`) with a never-acked PUBLISH and asserts `reportDroppedMsgs(1)` fires on the actual commit — not just the reflection shortcut into `reportSkippedMessagesIfAny`. It uses a real `ApplicationPersistedMsgCtx(Map.of(0L, 1), ...)` so the packetId is deterministic.

application/src/main/java/org/thingsboard/mqtt/broker/actors/client/state/PublishedInFlightCtxImpl.java
```diff
@@ -211,13 +218,24 @@ public void expireTtl(long ttlMs) {
             lock.unlock();
         }
 
+        // Persistence is a per-session invariant: check it once, not per message. A persistent-session copy is
+        // recoverable from the store so it never counts here; otherwise a non-retained message counts (see
+        // isCountableDrop for the full rule). Tally in the same pass that releases the buffers.
+        boolean nonPersistentSession = !clientSessionCtx.getSessionInfo().isPersistent();
+        int countableDrops = 0;
         for (MqttPublishMessage m : toRelease) {
+            if (nonPersistentSession && !m.fixedHeader().isRetain()) {
```
dmytro-landiak: This inlines the countable-drop rule (`nonPersistentSession && !isRetain()`) right next to `isCountableDrop()`, which still encodes the same rule and is still used by `addInFlightMsg`. The comment points a reader to `isCountableDrop`, but the two now have to be kept in sync by hand — and the NOTE just below even contemplates a QoS0 clause, so this is exactly the kind of predicate that could grow and get updated in one place but not the other. Since `getSessionInfo().isPersistent()` is just a field read, is the hoist mainly there to collapse the release loop and the old `stream().filter().count()` into a single pass? If so, you could keep one source of truth by having the loop call `isCountableDrop(m)` directly (or a small `isCountableDrop(nonPersistentSession, m)` overload that `addInFlightMsg` also feeds its freshly-read boolean to).

application/src/test/java/org/thingsboard/mqtt/broker/service/mqtt/delivery/DefaultMqttPublishMsgDeliveryServiceTest.java
```diff
@@ -161,6 +229,37 @@ public void givenStatsEnabled_whenSendAlreadyTracked_thenRecordsOnCompletion() t
         assertEquals(1, deliveryTimer().count());
     }
 
+    @Test
+    public void givenSyncThrowNonPersistentNonRetained_whenSendAlreadyTracked_thenReportsDroppedAndRecordsNothing() {
+        DefaultMqttPublishMsgDeliveryService service = buildService(true);
+        MqttFixedHeader fixedHeader = mock(MqttFixedHeader.class);
```
dmytro-landiak, line 235: The three-line fixed-header stub here (mock the header, stub `isRetain`, wire it onto `msg`) plus the optional `qosLevel` stub is now repeated across most of the drop tests in this file, and these two new tests add more copies. There's already a `stubPersistence(boolean)` helper following exactly this pattern — would a matching `stubFixedHeader(boolean retain)` / `stubFixedHeader(boolean retain, MqttQoS qos)` helper be worth adding, so each test reads as a line of intent rather than boilerplate?

application/src/test/java/org/thingsboard/mqtt/broker/service/mqtt/persistence/application/ApplicationPersistenceProcessorImplTest.java
```diff
@@ -487,6 +515,42 @@ void processMainPack_retryAll_whenClientNeverAcks_mustGiveUpAndCommitAfterRetryC
         verify(consumer, atLeastOnce()).commitSync();
     }
 
+    @Test
+    void processMainPack_retryAll_whenClientNeverAcksPublish_reportsDroppedOnGiveUpCommit() {
+        ApplicationAckStrategyConfiguration ackConfig = new ApplicationAckStrategyConfiguration();
+        ackConfig.setType(AckStrategyType.RETRY_ALL);
+        ackConfig.setRetries(2);
+        // Delegate to a real factory so the production RetryStrategy (with real give-up counting) is exercised.
+        ApplicationMsgAcknowledgeStrategyFactory realFactory = new ApplicationMsgAcknowledgeStrategyFactory(ackConfig);
```
dmytro-landiak, line 524: The first several setup lines here (real ack-strategy factory wiring, `packProcessingTimeout = 30L`, the `BurstSubmitStrategy` stub, `stopProcessorAfterDeliveries(20)`) are duplicated almost verbatim from `processMainPack_retryAll_whenClientNeverAcks_mustGiveUpAndCommitAfterRetryCap`, down to the identical `// Delegate to a real factory...` comment. Extracting a helper like `configureRealRetryStrategy(int retries)` would keep the two give-up tests focused on what actually differs (never-acked PUBREL vs never-acked PUBLISH) and stop the shared harness from drifting.

dmytro-landiak (commented):
## Re-review summary

Re-reviewed [feat(stats): complete droppedMsgs across the PUBLISH pipeline](https://github.com/thingsboard/tbmq/pull/346) — verified 4 finding(s) from [previous review](https://github.com/thingsboard/tbmq/pull/346#pullrequestreview-4664069207).

| Status | Count |
|--------|-------|
| ✅ Resolved | 3 |
| 💬 Acknowledged | 1 |
| ❌ Unresolved | 0 |

All previous findings have been addressed. Also found 3 new observations in the fix commits, commented inline — all non-blocking DRY/maintainability nits.

### Finding details

- ✅ **PublishedInFlightCtxImpl.java** — `isCountableDrop` comment claimed consistency with the delivery-service predicate when the two intentionally differ — *Fixed: comment reworded (a1f8dc9ea) to state the QoS0-clause omission is deliberate and warn against "aligning" them.*
- ✅ **DefaultMqttPublishMsgDeliveryService.java** — the sync-throw drop branch of `sendAlreadyTrackedPublishMsgToClient` was untested — *Fixed: two sync-throw tests added (7774bda90) covering the gate in both directions.*
- 💬 **ClientSessionCtx.java** — the 7-positional-arg `initPublishedInFlightCtx` has a swap-prone trailing int pair — *Developer: single compiler-checked construction site; the swap-prone int pair predates this PR; a param-object/builder is a standalone refactor, out of scope for a metrics-completeness PR. Reasonable.*
- ✅ **ApplicationPersistenceProcessorImplTest.java** — give-up counting was only exercised via reflection into the private `reportSkippedMessagesIfAny`, skipping the real commit gating — *Fixed: a test now drives the real RETRY_ALL give-up commit path with a never-acked PUBLISH and asserts `reportDroppedMsgs(1)` (7774bda90).*

---

PR #347, opened 2026-07-09
feat(server): enrich MqttSessionHandler.exceptionCaught logs with client id and close origin

dmytro-landiak:
## Pull Request description

`MqttSessionHandler.exceptionCaught` logged connection-level errors with only the Netty session id, e.g.:

```
WARN o.t.m.b.server.MqttSessionHandler - [1484fe0f-0603-40aa-8f83-0c76c8a33a96] IOException: Connection reset
```

That made recurring `IOException: Connection reset` warnings hard to act on: there was no way to tell **which MQTT client** was affected, nor whether the reset originated inside TBMQ or outside it.

This PR expands those logs:

- **Client id + remote address on every branch.** Every `exceptionCaught` branch now logs `[sessionId][clientId][remoteAddress]`. The address prefers the value captured on the first inbound bytes and falls back to `getAddress(ctx)`, so even pre-CONNECT failures (e.g. SSL handshake) still identify the peer. `clientId` is `null` until CONNECT is processed, which renders harmlessly.
- **Concrete exception class + best-effort close-origin verdict for IOExceptions.** The IOException branch now logs the actual exception type (e.g. `java.net.SocketException`) and attributes the close:
  - a new `closeInitiated` flag on `ClientSessionCtx` is set on every broker-initiated teardown — inside `closeChannel()`, the single chokepoint for graceful disconnects, keep-alive timeout, rate-limit/protocol errors and cluster takeover, plus the direct-close `connAckAndCloseCtx` auth-refusal path;
  - `exceptionCaught` reads it to state whether the connection was closed **by TBMQ** or **by the remote peer / a network device (external to TBMQ)**.

A `Connection reset` / `Broken pipe` IOException means TBMQ received a TCP RST or wrote to an already-closed socket, so it is never raised by TBMQ itself; the flag lets us tell a broker-driven teardown apart from a spontaneous peer/network abort. The attribution is intentionally **best-effort** — a reset arriving before the actor pipeline reaches `closeChannel()`, or a close driven by server shutdown, is not recorded and reads as external.

Resulting line:

```
WARN o.t.m.b.server.MqttSessionHandler - [1484fe0f-0603-40aa-8f83-0c76c8a33a96][my-device-01][/203.0.113.7:51234] IOException (java.net.SocketException): Connection reset. closed-by=peer-or-network
```

No behavior change beyond logging and the additive flag; the handler still ends in `disconnect(ON_ERROR, ...)`.

## General checklist

  - _Applied `Core` — this is a minor server-side/core logging improvement._
  - _No documentation changes required — this is an internal WARN log-format change with no user-facing, config or API surface._
  - _Fully backward compatible: log-message text only, plus one additive in-memory field on `ClientSessionCtx`; no schema, API or configuration changes._

  - _N/A — no front-end changes._

## Back-End feature checklist

  - _`ClientSessionCtxTest` covers the `closeInitiated` lifecycle (defaults false; set by `closeChannel()`), and `MqttSessionHandlerTest` covers both branches of the close-origin text. The full `exceptionCaught` path is not unit-tested because `MqttSessionHandler` is tightly coupled to the Netty channel and `MqttHandlerCtx` services; the origin logic was extracted to a static `connectionCloseOrigin(boolean)` helper to make it directly testable, and the rendered log lines were verified out-of-band via SLF4J's `MessageFormatter`._
  - _No new dependencies._

### DIFF
--- a/application/src/main/java/org/thingsboard/mqtt/broker/server/MqttSessionHandler.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/server/MqttSessionHandler.java
@@ -345,7 +345,7 @@ public class MqttSessionHandler extends ChannelInboundHandlerAdapter implements 
             log.warn("[{}][{}][{}] NotSslRecordException: {}", sessionId, clientId, remoteAddress, cause.getCause().getMessage());
             exceptionMessage = cause.getCause().getMessage();
         } else if (cause instanceof IOException) {
-            log.warn("[{}][{}][{}] IOException ({}): {}. Connection closed {}.",
+            log.warn("[{}][{}][{}] IOException ({}): {}. closed-by={}",
                     sessionId, clientId, remoteAddress, cause.getClass().getName(), cause.getMessage(),
                     connectionCloseOrigin(clientSessionCtx.isCloseInitiated()));
             exceptionMessage = cause.getMessage();
@@ -364,18 +364,18 @@ public class MqttSessionHandler extends ChannelInboundHandlerAdapter implements 
      * (typically "Connection reset" / "Connection reset by peer" / "Broken pipe"). Such an error means TBMQ
      * received a TCP RST or wrote to an already-closed socket, so it is never raised by TBMQ itself. We use
      * whether a broker-side close was recorded for this session (see {@link ClientSessionCtx#isCloseInitiated()})
-     * to tell the two situations apart:
+     * to tell the two situations apart and return a short, stable {@code closed-by} token for the log line:
      * <ul>
-     *   <li>recorded — the reset is part of a teardown TBMQ started (rate limit, protocol error, takeover, ...);</li>
-     *   <li>not recorded — the client or a network device between the client and TBMQ aborted the connection.</li>
+     *   <li>{@code "TBMQ"} — a broker-side close was recorded, so the reset is part of a teardown TBMQ started
+     *       (rate limit, protocol error, takeover, ...);</li>
+     *   <li>{@code "peer-or-network"} — no broker-side close was recorded, so the client or a network device
+     *       between the client and TBMQ aborted the connection (external to TBMQ).</li>
      * </ul>
      * It is best-effort: a reset arriving before the actor pipeline reaches {@code closeChannel()}, or a close
-     * driven by server shutdown, is not recorded and therefore reads as external.
+     * driven by server shutdown, is not recorded and therefore reads as {@code peer-or-network}.
      */
     static String connectionCloseOrigin(boolean brokerCloseRecorded) {
-        return brokerCloseRecorded
-                ? "by TBMQ (a broker-side close was recorded for this session; the peer's reset is part of that teardown)"
-                : "by the remote peer or a network device between the client and TBMQ (external to TBMQ; no broker-side close was recorded for this session)";
+        return brokerCloseRecorded ? "TBMQ" : "peer-or-network";
     }
 
     @Override
--- a/application/src/main/java/org/thingsboard/mqtt/broker/server/MqttSessionHandler.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/server/MqttSessionHandler.java
@@ -222,11 +222,10 @@ public class MqttSessionHandler extends ChannelInboundHandlerAdapter implements 
 
     private void connAckAndCloseCtx(MqttConnectReturnCode reasonCode) {
         var mqttConnAckMessage = mqttMessageGenerator.createMqttConnAckMsg(reasonCode);
-        // Broker-initiated close: record it so exceptionCaught can attribute a following reset to TBMQ rather
-        // than the peer. This path closes the channel directly instead of going through ClientSessionCtx#closeChannel.
-        clientSessionCtx.setCloseInitiated(true);
         clientSessionCtx.getChannel().writeAndFlush(mqttConnAckMessage);
-        clientSessionCtx.getChannel().close();
+        // Funnel this broker-initiated close through the single chokepoint (which records it via closeInitiated,
+        // read by exceptionCaught) rather than closing the channel directly.
+        clientSessionCtx.closeChannel();
     }
 
     private void processPublish(MqttMessage msg) {
--- a/application/src/main/java/org/thingsboard/mqtt/broker/server/MqttSessionHandler.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/server/MqttSessionHandler.java
@@ -222,6 +222,9 @@ public class MqttSessionHandler extends ChannelInboundHandlerAdapter implements 
 
     private void connAckAndCloseCtx(MqttConnectReturnCode reasonCode) {
         var mqttConnAckMessage = mqttMessageGenerator.createMqttConnAckMsg(reasonCode);
+        // Broker-initiated close: record it so exceptionCaught can attribute a following reset to TBMQ rather
+        // than the peer. This path closes the channel directly instead of going through ClientSessionCtx#closeChannel.
+        clientSessionCtx.setCloseInitiated(true);
         clientSessionCtx.getChannel().writeAndFlush(mqttConnAckMessage);
         clientSessionCtx.getChannel().close();
     }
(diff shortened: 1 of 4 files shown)


application/src/main/java/org/thingsboard/mqtt/broker/server/MqttSessionHandler.java
```diff
@@ -332,26 +335,50 @@ public void channelReadComplete(ChannelHandlerContext ctx) {
...
+     * <ul>
+     *   <li>recorded — the reset is part of a teardown TBMQ started (rate limit, protocol error, takeover, ...);</li>
+     *   <li>not recorded — the client or a network device between the client and TBMQ aborted the connection.</li>
+     * </ul>
+     * It is best-effort: a reset arriving before the actor pipeline reaches {@code closeChannel()}, or a close
+     * driven by server shutdown, is not recorded and therefore reads as external.
+     */
+    static String connectionCloseOrigin(boolean brokerCloseRecorded) {
+        return brokerCloseRecorded
+                ? "by TBMQ (a broker-side close was recorded for this session; the peer's reset is part of that teardown)"
```
dmytro-landiak: This verdict is a full explanatory sentence (~100 chars), and per the PR description it fires on the recurring `Connection reset` path — so every reset now emits a long, prose-heavy WARN line. The method name promises an *origin* (a short classification) but returns a paragraph. Would a concise token read better here — e.g. `closed-by=TBMQ` / `closed-by=peer-or-network` — keeping the rationale where it already lives, in the Javadoc? That also removes the triple redundancy of the same explanation appearing in the Javadoc, both return strings, and the test comments.

application/src/main/java/org/thingsboard/mqtt/broker/server/MqttSessionHandler.java
```diff
@@ -332,26 +335,50 @@ public void channelReadComplete(ChannelHandlerContext ctx) {
 
     @Override
     public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
+        // Prefer the address captured on the first inbound bytes; fall back to the channel so pre-CONNECT
+        // failures (e.g. SSL handshake) still identify the remote peer. clientId may be null before CONNECT.
+        InetSocketAddress remoteAddress = address != null ? address : getAddress(ctx);
```
dmytro-landiak, line 337: The genuinely new behavior in this PR isn't covered by tests: the enriched branch logging, this `address != null ? address : getAddress(ctx)` fallback (a real two-outcome branch), and `connAckAndCloseCtx` now setting `closeInitiated`. The only new `MqttSessionHandler` test exercises the pure `connectionCloseOrigin` string helper — the least risky part. Could we add a test — integration if feasible, otherwise a handler-level unit test with a mocked `ctx`/channel — that drives `exceptionCaught` with and without a captured address and verifies `connAckAndCloseCtx` records the close, so the attribution path is actually exercised? I see the PR notes the full path is hard to unit-test due to the Netty coupling; even asserting the `connAckAndCloseCtx` flag-set (via `ClientSessionCtx`) would close the biggest gap.

application/src/test/java/org/thingsboard/mqtt/broker/server/MqttSessionHandlerTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class MqttSessionHandlerTest {
+
+    // When a broker-side close was recorded for the session, a following IOException/reset is attributed to TBMQ.
+    @Test
+    public void connectionCloseOrigin_brokerCloseRecorded_attributesToTbmq() {
+        assertThat(MqttSessionHandler.connectionCloseOrigin(true))
+                .contains("by TBMQ")
```
dmytro-landiak: These assertions pin substrings of a human-readable log sentence, so any rewording of the message breaks the test even though behavior is unchanged — it's effectively a change-detector that restates the two-constant selector. If the verdict were reduced to a short stable token/enum (see the related comment on `connectionCloseOrigin`), the assertion could target that instead and stop tracking the exact wording.

application/src/main/java/org/thingsboard/mqtt/broker/server/MqttSessionHandler.java
```diff
@@ -222,6 +222,9 @@ private void processMqttMsg(MqttMessage msg) {
 
     private void connAckAndCloseCtx(MqttConnectReturnCode reasonCode) {
         var mqttConnAckMessage = mqttMessageGenerator.createMqttConnAckMsg(reasonCode);
+        // Broker-initiated close: record it so exceptionCaught can attribute a following reset to TBMQ rather
+        // than the peer. This path closes the channel directly instead of going through ClientSessionCtx#closeChannel.
+        clientSessionCtx.setCloseInitiated(true);
```
dmytro-landiak: The invariant "every broker-initiated close records `closeInitiated`" is now enforced by convention in two separate spots — `ClientSessionCtx.closeChannel()` and this path, which deliberately bypasses it. A future broker-side close path that forgets the flag will silently misattribute resets to the peer, and `ClientSessionCtxTest` even documents `closeChannel()` as "the single chokepoint," which this path contradicts. Could the flag live in one place — e.g. route `connAckAndCloseCtx` through `closeChannel()` after writing the CONNACK, or set the flag wherever the channel is actually closed — so the two can't drift?

dmytro-landiak (commented):
## Review summary

Reviewed 4 changed files in [feat(server): enrich MqttSessionHandler.exceptionCaught logs with client id and close origin](https://github.com/thingsboard/tbmq/pull/347). Left 4 comment(s) inline.

No correctness issues found — this is a well-scoped logging enhancement plus one additive `volatile` field. `clientSessionCtx` is constructed in the handler's constructor (never null in `exceptionCaught`), a null `clientId`/`remoteAddress` renders harmlessly, the flag is per-connection so it can't leak across connections, and the cross-thread read/write is correctly `volatile` (the `address` field is only touched on the channel's single event-loop thread). The inline comments are all quality/design suggestions: log-line verbosity, the `closeInitiated` invariant being split across two close paths, and test coverage of the genuinely new behavior.

---

application/src/main/java/org/thingsboard/mqtt/broker/server/MqttSessionHandler.java
dmytro-landiak: Done in 927d77fc5. `connectionCloseOrigin` now returns a short, stable token — `TBMQ` / `peer-or-network` — and the WARN line reads `... closed-by={}`, e.g. `closed-by=peer-or-network` (greppable). The rationale (best-effort attribution, TCP-RST semantics) now lives only in the method Javadoc, removing the triple redundancy.

application/src/test/java/org/thingsboard/mqtt/broker/server/MqttSessionHandlerTest.java
dmytro-landiak: Fixed in 927d77fc5. With the verdict reduced to a stable token, the assertions now target the value directly — `isEqualTo("TBMQ")` / `isEqualTo("peer-or-network")` — so they no longer track the exact wording of a human-readable sentence.

application/src/main/java/org/thingsboard/mqtt/broker/server/MqttSessionHandler.java
dmytro-landiak: Added handler-level tests in 927d77fc5 (mocked `ChannelHandlerContext`/`Channel`): (1) `getAddress` returns the channel attribute when present and falls back to `remoteAddress()` when absent — the two-outcome branch behind `address != null ? address : getAddress(ctx)`; (2) the enriched IOException branch, driven through `exceptionCaught`, asserting (via a `ListAppender`) the WARN line carries the exception class, the message and `closed-by=peer-or-network`, with the address resolved through the fallback; (3) `connAckAndCloseCtx` driven through `channelRead` with an unsupported auth method, verifying it writes the CONNACK and then closes through `ClientSessionCtx.closeChannel()` (i.e. records the broker-side close).

application/src/main/java/org/thingsboard/mqtt/broker/server/MqttSessionHandler.java
dmytro-landiak: Resolved in e990aedbb: `connAckAndCloseCtx` now writes the CONNACK and then calls `clientSessionCtx.closeChannel()` instead of setting the flag by hand and closing the channel directly. The `closeInitiated` invariant now lives in exactly one place — `closeChannel()` — so the two paths can't drift, and `ClientSessionCtxTest`'s "single chokepoint" description holds. 927d77fc5 adds a test that drives this path and asserts the CONNACK-then-close.

PR #348, opened 2026-07-09
feat(stats): connection-outcome metrics (accepted/refused/error)

dmytro-landiak:
## Pull Request description

Adds three **additive, untagged** Prometheus counters (emitted when `stats.enabled=true`) covering every terminal outcome of an inbound MQTT connection attempt:

| Meter (scraped) | Counts |
|---|---|
| `connectionAccepted_total` | successful CONNACKs |
| `connectionRefused_total` | refusal CONNACKs (any `CONNECTION_REFUSED_*` code) |
| `connectionError_total` | establishment failures with no CONNACK (TLS handshake / non-SSL framing / IO / protocol violation) |

**Scope of changes:**
- **Where counted:** accept/refuse at the two `DefaultMqttMessageCreator.createMqttConnAckMsg` CONNACK choke points (the single place all CONNACKs are built, shared by all transports); establishment errors at `MqttSessionHandler.exceptionCaught`, gated on `clientId == null`. The single-arg `createMqttConnAckMsg` classifies by the **actual return code** (`CONNECTION_ACCEPTED` → accepted, otherwise refused), so the outcome is driven by the code rather than by which overload a caller happens to use.
- **Seam:** a new `ConnectionStats` facade vended by `StatsManager` under the existing `stats.enabled` master switch — interface + `DefaultConnectionStats` (three counters) + no-op `StubConnectionStats`. Threaded to the per-connection handler via `MqttHandlerCtx`.
- **Untagged, no per-cause discriminator.** All three methods — `onConnectionAccepted()` / `onConnectionRefused()` / `onConnectionError()` — are no-arg; each counter is a single untagged series, and `rate()`/`increase()` give the accept/refuse/error throughput. Per-cause label breakdowns (by refusal return code or error type) are intentionally **not** part of this community counter set (YAGNI) — nothing is threaded through the seam for them.
- **Disjointness / lower bound:** the `clientId == null` gate keeps `connectionError` disjoint from the existing `clientDisconnects` counter (post-session channel errors). Consequently `connectionAccepted + connectionRefused + connectionError` (derived in PromQL) is a **lower bound** on total attempts, not an exact identity — a channel error after CONNECT but before the session is established is counted by neither family. This is intentional (to keep the two families disjoint) and documented in the `ConnectionStats` Javadoc (the single source of truth) and at the `exceptionCaught` gate.
- **Periodic log:** the three interval counts are logged together on one `[connection] Stats: …` line.

**Documentation notes:** these are runtime meters, not configuration keys, so no yml key is added and no config/migration changes are required. Public metrics documentation should list `connectionAccepted_total`, `connectionRefused_total`, and `connectionError_total` (all untagged), note that `total = accepted + refused + error` is a **lower bound** on connection attempts, and note that per-cause label breakdowns are not part of this community counter set.

> `develop/2.4` has been merged into this branch; one add/add conflict in `MqttSessionHandlerTest` (both branches added the class) was resolved by combining develop's close-origin/`getAddress`/enriched-WARN suite with the connection-error tests.

## General checklist

- [x] Screenshots with affected component(s) are added. The best option is to provide 2 screens: before and after changes; _(N/A — back-end only, no UI changes)_
- [x] If you change the widget or other API, ensure it is backward-compatible or upgrade script is present. _(N/A — back-end only)_

## Back-End feature checklist

- [x] Added corresponding unit and/or integration test(s). Provide written explanation in the PR description if you have failed to add tests. _(`DefaultConnectionStatsTest`, connection-error cases in the merged `MqttSessionHandlerTest`, and new cases in `DefaultMqttMessageCreatorTest`; regression `StatsManagerImplTest`/`ConnectServiceImplTest`/`ActorProcessorImplTest` still green — 68 tests across the affected set, 0 failures, Java 25.)_
- [x] If new dependency was added: the dependency tree is checked for conflicts. _(N/A — no new dependencies.)_


application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/DefaultMqttMessageCreator.java
```diff
@@ -103,11 +111,15 @@ public void init() {
 
     @Override
     public MqttConnAckMessage createMqttConnAckMsg(MqttConnectReturnCode returnCode) {
+        // This single-arg overload is the refusal-only CONNACK choke point (accepts use the two-arg
+        // overload), so every call here is a refused connection.
+        connectionStats.onConnectionRefused(returnCode);
```
dmytro-landiak: The refusal count keys off *which overload* was called rather than the actual `returnCode` — the comment above notes every call here is a refusal. I traced all current callers of this single-arg overload (`MqttSessionHandler.connAckAndCloseCtx`, `ActorProcessorImpl.sendConnectionRefusedMsgAndCloseChannel`, `ConnectServiceImpl.createAndSendConnAckMsg`) and they do all pass `CONNECTION_REFUSED_*` codes, so it's correct today. But it's a fragile coupling: `CONNECTION_ACCEPTED` is a legal argument to this signature, and a future caller that routes an accept through here would be silently miscounted as a refusal with nothing to catch it. Would guarding the increment on `returnCode != CONNECTION_ACCEPTED` make the classification robust to the code itself rather than relying on caller discipline across the codebase?

application/src/main/java/org/thingsboard/mqtt/broker/service/stats/ConnectionStats.java
```diff
@@ -0,0 +1,66 @@
...
+public interface ConnectionStats {
+
+    /** Records one accepted connection (successful CONNACK). */
+    void onConnectionAccepted();
+
+    /**
+     * Records one refused connection (a {@code CONNECTION_REFUSED_*} CONNACK). {@code returnCode} is
+     * carried so the PE extension can tag by reason; the CE impl ignores it (single untagged counter).
+     */
+    void onConnectionRefused(MqttConnectReturnCode returnCode);
```
dmytro-landiak: Slight asymmetry on the seam: the *error* discriminator got a new broker-internal enum (`ConnectionErrorType`), and the sibling `ClientDisconnectStats` takes the internal `DisconnectReasonType`, but the *refusal* discriminator here is the raw Netty `MqttConnectReturnCode`. It's defensible — `MqttConnectReturnCode` is already the broker's vocabulary for connect return codes across `ConnectServiceImpl`/`ActorProcessorImpl`, whereas no domain type existed for pre-CONNACK failures (hence `ConnectionErrorType`). Still worth a conscious call: either give the refusal reason a domain enum too for consistency and to keep the stats seam decoupled from the Netty codec, or — if leaking the Netty type is acceptable here — `ConnectionErrorType` is arguably a thin wrapper over the exception classification. Was the split intentional?

application/src/main/resources/thingsboard-mqtt-broker.yml
```diff
@@ -1169,6 +1169,14 @@ redis:
     # Enables or disables blocking when a thread requests a connection from an exhausted pool.
     blockWhenExhausted: "${REDIS_POOL_CONFIG_BLOCK_WHEN_EXHAUSTED:true}"
 
+# Connection-outcome Prometheus counters (emitted when stats.enabled=true), untagged in CE:
+#   connectionAccepted_total  - successful CONNACKs.
+#   connectionRefused_total   - refused CONNACKs (any CONNECTION_REFUSED_* code).
+#   connectionError_total     - connection-establishment failures with no CONNACK (TLS/framing/IO).
+# accepted + refused + error (derived in PromQL) is a lower bound on total attempts: an error that
```
dmytro-landiak: This nuanced invariant — accepted+refused+error is a *lower bound*, and the two metric families are deliberately kept disjoint — is written out verbatim in three places: this yml comment, the `ConnectionStats` interface Javadoc, and the inline comment in `MqttSessionHandler.exceptionCaught`. If the gating logic ever changes, all three have to move in lockstep or they'll drift and mislead. Might be worth making one the source of truth (the interface Javadoc seems the natural home) and having the yml and the inline comment point to it with a one-liner rather than restating the whole rationale.

application/src/test/java/org/thingsboard/mqtt/broker/server/MqttSessionHandlerTest.java
```diff
@@ -0,0 +1,97 @@
...
+    public void givenNoClientId_whenUnknownException_thenConnectionErrorOtherCounted() {
+        handler.exceptionCaught(mock(ChannelHandlerContext.class), new RuntimeException("boom"));
+        verify(connectionStats).onConnectionError(ConnectionErrorType.OTHER);
+    }
+
+    @Test
+    public void givenEstablishedClientId_whenException_thenConnectionErrorNotCounted() {
+        // A post-session channel error (clientId != null) is a disconnect (clientDisconnects), not a
+        // connection-establishment error — connectionError must NOT be counted.
+        ReflectionTestUtils.setField(handler, "clientId", "client-a");
```
dmytro-landiak, line 212: The three new behaviors — accept-path counting, refusal counting, and this `clientId == null` establishment-error gate — are all verified against a mocked `ConnectionStats` (and the gate specifically via reflection to force `clientId`). The per-branch unit coverage is genuinely thorough. What nothing exercises is the real flow end to end: that these CONNACK choke points and `exceptionCaught` are actually the paths hit on a real connect / refuse / TLS-handshake failure, and that the counters are wired and exposed under `stats.enabled`. So this is more a wiring-confidence gap than a branch gap — but given TBMQ already has connection-level integration test infrastructure, an integration test asserting a counter moves on a real connection outcome would be stronger. Trade-off acknowledged: asserting on Prometheus counters end to end is heavier and a bit more brittle than these focused unit tests.

dmytro-landiak (commented):
## Review summary

Reviewed 15 changed files in [feat(stats): connection-outcome metrics (accepted/refused/error)](https://github.com/thingsboard/tbmq/pull/348). Left 4 comment(s) inline.

The change is clean on the correctness front: the Spring bean lifecycle is safe (no cycle, so `StatsManagerImpl.init()` runs before `DefaultMqttMessageCreator`'s constructor reads `getConnectionStats()`), the `clientId == null` gate is correct and thread-confined to the Netty event loop, `DefaultCounter` is thread-safe, and `reset()` correctly keeps the Prometheus `_total` counters monotonic. The disjointness between `connectionError` and `clientDisconnects` holds end to end. The inline comments are maintainability / design / test-coverage observations, not bugs.

---

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/DefaultMqttMessageCreator.java
```diff
@@ -103,11 +111,21 @@ public void init() {
 
     @Override
     public MqttConnAckMessage createMqttConnAckMsg(MqttConnectReturnCode returnCode) {
+        // This single-arg overload is the refusal path (accepts normally use the two-arg overload), but
+        // classify by the actual return code rather than by which overload was called: CONNECTION_ACCEPTED
+        // is a legal argument to this signature, so guarding on the code keeps the count correct even if a
+        // future caller routes an accept through here, instead of silently miscounting it as a refusal.
+        if (returnCode == CONNECTION_ACCEPTED) {
```
dmytro-landiak, line 118: This accept-branch (and its four-line justification) defends against a caller that routes an accept through the single-arg overload — a path that doesn't exist today. Worth weighing against YAGNI: if an accept ever did come through here it would build a bare CONNACK missing every MQTT 5 property the two-arg overload adds (assigned client id, keep-alive, receive-max, session-expiry…), so the connection would be broken regardless of which counter incremented. Given that, is guarding the stat's accuracy on an already-malformed path pulling its weight, or would a one-line comment documenting the refusal-only invariant convey the intent with less machinery? It also means accept-counting now lives in two spots (`onConnectionAccepted()` here and in the two-arg overload) that have to stay in sync. Not a strong objection — just flagging the trade-off, since this branch grew out of last round's suggestion to classify by return code.

application/src/main/java/org/thingsboard/mqtt/broker/server/MqttSessionHandler.java
```diff
@@ -354,6 +357,14 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
             log.error("[{}][{}][{}] Unexpected Exception", sessionId, clientId, remoteAddress, cause);
             exceptionMessage = cause.getMessage();
         }
+        // Count only pre-establishment errors: clientId is still null, i.e. no CONNECT has been processed
+        // yet (no session). Post-session channel errors (clientId != null) are counted as clientDisconnects
+        // via disconnect(ON_ERROR) below. This gate deliberately leaves connectionError a lower bound: an
```
dmytro-landiak, line 362: Removing the yml copy addressed the main duplication concern from the last round. One lighter residual: the lower-bound / disjoint-families rationale is still stated almost verbatim both here and in the `ConnectionStats` Javadoc, so the two have to be edited in lockstep or they drift. They do serve slightly different readers (local gate reasoning vs. metric semantics for an API consumer), so I wouldn't merge them — but would it be cleaner to keep the full explanation in the interface Javadoc and trim this inline block to a one-liner plus a pointer (e.g. "gate on clientId == null keeps connectionError a lower bound and disjoint from clientDisconnects — see ConnectionStats javadoc")? Optional polish.

application/src/main/java/org/thingsboard/mqtt/broker/service/stats/ConnectionStats.java
```diff
@@ -0,0 +1,64 @@
...
```
dmytro-landiak: Now that `ConnectionErrorType` is gone, the two failure methods have inconsistent shapes: `onConnectionRefused` still carries a discriminator the CE impl ignores (a hook so PE can tag by return code), while `onConnectionError()` carries none. The Javadoc says that's intentional — `connectionError` stays untagged in both editions — so this looks settled. Flagging only for the record: dropping the enum also closed PE's extension point for breaking errors down by cause (SSL handshake vs. IO vs. framing, which the deleted enum captured right at the call site), so reintroducing that later would be an interface signature change rather than a pure override. If an error-cause breakdown is firmly out of scope for PE too, no action needed.

dmytro-landiak (commented):
## Re-review summary

Re-reviewed [feat(stats): connection-outcome metrics (accepted/refused/error)](https://github.com/thingsboard/tbmq/pull/348) — verified 4 finding(s) from [previous review](https://github.com/thingsboard/tbmq/pull/348#pullrequestreview-4664515194).

| Status | Count |
|--------|-------|
| ✅ Resolved | 3 |
| 💬 Acknowledged | 0 |
| ❌ Unresolved | 1 |

Also found 3 quality/design observation(s) on the fix commits, commented inline.

### Finding details

- ✅ **DefaultMqttMessageCreator.java:118** — single-arg CONNACK was classified by which overload was called rather than by the actual return code — *Fixed: `CONNECTION_ACCEPTED` now routes to `onConnectionAccepted()`, all other codes to `onConnectionRefused()`.*
- ✅ **ConnectionStats.java:51** — refusal/error discriminator asymmetry (`ConnectionErrorType` enum vs. raw Netty `MqttConnectReturnCode`) — *Fixed: `ConnectionErrorType` was dropped, `onConnectionError()` is now untagged in both editions, and the choice is documented in the interface Javadoc.*
- ✅ **thingsboard-mqtt-broker.yml** — lower-bound rationale duplicated in 3 places (yml + Javadoc + inline gate comment) — *Fixed: the yml comment copy (the most drift-prone) was removed; the rationale now lives in the interface Javadoc and the inline gate comment.*
- ❌ **MqttSessionHandlerTest.java** — no end-to-end test that a counter actually moves on a real connect / refuse / handshake-failure outcome — *Still verified only via a mocked `ConnectionStats` in focused unit tests; no integration test was added and there was no reply. Low priority — the original finding itself acknowledged the added brittleness/cost trade-off, and the fix commits did add solid per-branch unit coverage (including the new single-arg accept branch).*

One documentation heads-up (not a code issue): the PR description's *Documentation notes* still say the rationale is documented via a comment block beside the `stats:` section of `thingsboard-mqtt-broker.yml`, but that block was removed in the fix commits — the `ConnectionStats` interface Javadoc is now the source of truth for the metric semantics, so public-docs work should key off that.

---

PR #349, opened 2026-07-10
Outbound rate limiting for persistent APPLICATION clients

### LINKED ISSUE #305: [Feature Request] Outbound rate limiting / throttling for MQTT subscribers (APPLICATION clients)
**Is your feature request related to a problem? Please describe.**

In our setup, we use MQTT (no Kafka consumer), QoS 1, and require full message retention (no data loss). When a subscriber (customer application) is offline for an extended period, messages accumulate in TBMQ (Kafka-backed persistence). Once the subscriber reconnects, TBMQ replays the backlog at maximum speed, overwhelming the subscriber. Since we cannot control or modify the subscriber application, which immediately ACKs messages and pushes them into an internal asyncio.Queue, the application’s memory usage increases rapidly, eventually causing it to crash.

**Describe the solution you'd like**

We would like to see outbound rate limiting / throttling for APPLICATION clients at the broker level. Specifically, we would like to be able to:

- Limit message delivery rate per subscriber (e.g., max 100 messages/sec)
- Throttle backlog replay to prevent bulk delivery (currently, delivery is happening at full speed, causing bursts)
- Allow configuring rate-limiting policies directly on TBMQ (not relying on subscriber-side changes)
- Smooth the backlog replay when the subscriber is recovering from being offline, ensuring that messages are delivered in steady chunks instead of all at once

**Describe alternatives you've considered**

We’ve explored the following broker-side configurations to manage the rate of message delivery:

- Netty backpressure tuning (NETTY_WRITE_BUFFER_LOW_WATER_MARK / NETTY_WRITE_BUFFER_HIGH_WATER_MARK) → helps reduce burst size but does not enforce a fixed rate
- Message delivery strategies (writeAndFlush vs. buffered) → prevents large batches but still does not limit the delivery rate
- Kafka consumer tuning (max.poll.records) → reduces batch size but does not directly limit the rate of delivery to subscribers
- Inflight message limits (max_inflight_messages) → forces PUBACK pacing but does not control message rate per second
- Queue limits (e.g., TB_APP_PERSISTED_MSG_PACK_PROCESSING_TIMEOUT) → introduces delays but still not true rate limiting

These solutions mitigate the issue but do not provide a rate-limiting mechanism at the broker level to control the number of messages sent per second.

**Describe the specific rate limit configuration you would like**

We would like the ability to set a maximum outbound message rate per client or per topic. For example:

- **Max messages per second** (e.g., 100 messages/sec per subscriber)
- **Max messages per minute** (e.g., 5000 messages/minute)

This would allow us to avoid flooding the subscriber with a large volume of messages and control the rate of delivery during backlog replay.

**Additional context**

- Subscriber uses Python aiomqtt, where ACK is automatic and cannot be delayed or controlled
- Messages are considered delivered by TBMQ as soon as received, even if internally queued
- Current delivery model is effectively batch-driven (Kafka poll → send → wait → repeat), leading to bursty behavior
- A true rate limiter at broker level would significantly improve stability for such clients

dmytro-landiak:
## Pull Request description

Closes #305.

Adds an **opt-in, per-APPLICATION-client outbound rate limit** that *paces* (throttles) delivery of persisted messages instead of dropping them. When a persistent APPLICATION subscriber reconnects after being offline, its Kafka backlog is currently replayed at full speed, which can overwhelm subscribers that auto-ACK and buffer internally. This throttles delivery to a configured rate, smoothing the replay into steady chunks. **No data loss**: offsets are committed only after acks and the backlog stays in Kafka — the per-client consumer thread is simply paced.

**Scope of changes:**
- New opt-in config `mqtt.rate-limits.application-persisted-messages` (`enabled` + `client-config`, same `limit:seconds,...` format as the other limiters; default **off**). Env vars `MQTT_APPLICATION_PERSISTED_MSGS_RATE_LIMITS_ENABLED` / `..._CLIENT_CONFIG`.
- `RateLimitService`: per-client `TbRateLimits` bucket (reusing the existing bucket4j framework) with `tryConsumeApplicationPersistedMsgs` / `isApplicationPersistedMsgsRateLimitEnabled`; cleanup rides the existing `remove(clientId)` on disconnect.
- `ApplicationPersistenceProcessorImpl`: a session-aware throttle gate applied once per pack before delivery (in both the main and shared-subscription loops, before the retry loop so QoS retransmissions don't double-count; counts PUBLISH only).

**Documentation notes:** document the new setting/env vars and the shared-subscription caveat (avoid a rate so low that draining one polled pack takes longer than `max.poll.interval.ms`, or the shared consumer group may rebalance — the per-client main consumer uses manual partition assignment and is unaffected). A follow-up ticket will add a safeguard for that shared-subscription edge case.

## General checklist

- [ ] You have reviewed the guidelines [document](https://docs.google.com/document/d/1A4Vc3wrHsY_159b9RG5LOtCryoH6VPwgOhrFEzJKwbI/edit?usp=sharing).
- [ ] [Labels](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels#about-labels) that classify your pull request have been added.
- [ ] The [milestone](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/about-milestones) is specified and corresponds to fix version.
- [x] Description references specific [issue](https://github.com/thingsboard/tbmq/issues). (#305)
- [x] Changes are backward compatible or upgrade script is provided. (opt-in, default off)

- [x] N/A — no front-end changes.

## Back-End feature checklist

- [x] Added corresponding unit and/or integration test(s). (`RateLimitServiceImplTest`, `ApplicationPersistenceProcessorImplTest`)
- [x] If new dependency was added: the dependency tree is checked for conflicts. (no new dependency)

### DIFF
--- a/application/src/main/java/org/thingsboard/mqtt/broker/service/limits/RateLimitService.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/service/limits/RateLimitService.java
@@ -49,7 +49,7 @@ public interface RateLimitService {
     /**
      * Non-blocking token consume for outbound persisted messages to an APPLICATION client (throttling, not dropping)
      */
-    boolean tryConsumeApplicationPersistedMsg(String clientId);
+    boolean tryConsumeApplicationPersistedMsgs(String clientId);
 
     boolean isApplicationPersistedMsgsRateLimitEnabled();
 
--- a/application/src/main/java/org/thingsboard/mqtt/broker/service/limits/RateLimitService.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/service/limits/RateLimitService.java
@@ -46,6 +46,13 @@ public interface RateLimitService {
 
     boolean isDevicePersistedMsgsLimitEnabled();
 
+    /**
+     * Non-blocking token consume for outbound persisted messages to an APPLICATION client (throttling, not dropping)
+     */
+    boolean tryConsumeApplicationPersistedMsg(String clientId);
+
+    boolean isApplicationPersistedMsgsRateLimitEnabled();
+
     long tryConsumeTotalMsgs(long limit);
 
     boolean isTotalMsgsLimitEnabled();
--- a/application/src/main/java/org/thingsboard/mqtt/broker/service/limits/RateLimitServiceImpl.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/service/limits/RateLimitServiceImpl.java
@@ -24,6 +24,7 @@ import org.springframework.stereotype.Service;
 import org.thingsboard.mqtt.broker.common.data.ClientSessionInfo;
 import org.thingsboard.mqtt.broker.common.data.SessionInfo;
 import org.thingsboard.mqtt.broker.common.util.TbRateLimits;
+import org.thingsboard.mqtt.broker.config.ApplicationPersistedMsgsRateLimitsConfiguration;
 import org.thingsboard.mqtt.broker.config.ClientsLimitProperties;
 import org.thingsboard.mqtt.broker.config.DevicePersistedMsgsRateLimitsConfiguration;
 import org.thingsboard.mqtt.broker.config.IncomingRateLimitsConfiguration;
@@ -44,6 +45,7 @@ public class RateLimitServiceImpl implements RateLimitService {
     private final OutgoingRateLimitsConfiguration outgoingRateLimitsConfiguration;
     private final TotalMsgsRateLimitsConfiguration totalMsgsRateLimitsConfiguration;
     private final DevicePersistedMsgsRateLimitsConfiguration devicePersistedMsgsRateLimitsConfiguration;
+    private final ApplicationPersistedMsgsRateLimitsConfiguration applicationPersistedMsgsRateLimitsConfiguration;
     private final RateLimitCacheService rateLimitCacheService;
     private final ClientsLimitProperties clientsLimitProperties;
 
@@ -51,6 +53,8 @@ public class RateLimitServiceImpl implements RateLimitService {
     private ConcurrentMap<String, TbRateLimits> incomingPublishClientLimits;
     @Getter
     private ConcurrentMap<String, TbRateLimits> outgoingPublishClientLimits;
+    @Getter
+    private ConcurrentMap<String, TbRateLimits> applicationPersistedMsgClientLimits;
 
     @PostConstruct
     public void init() {
@@ -59,6 +63,9 @@ public class RateLimitServiceImpl implements RateLimitService {
         }
         if (outgoingRateLimitsConfiguration.isEnabled()) {
             outgoingPublishClientLimits = new ConcurrentHashMap<>();
+        }
+        if (applicationPersistedMsgsRateLimitsConfiguration.isEnabled()) {
+            applicationPersistedMsgClientLimits = new ConcurrentHashMap<>();
         }
     }
 
@@ -99,6 +106,9 @@ public class RateLimitServiceImpl implements RateLimitService {
             }
             if (outgoingPublishClientLimits != null) {
                 outgoingPublishClientLimits.remove(clientId);
+            }
+            if (applicationPersistedMsgClientLimits != null) {
+                applicationPersistedMsgClientLimits.remove(clientId);
             }
         }
     }
@@ -173,6 +183,21 @@ public class RateLimitServiceImpl implements RateLimitService {
     @Override
     public boolean isDevicePersistedMsgsLimitEnabled() {
         return devicePersistedMsgsRateLimitsConfiguration.isEnabled();
+    }
+
+    @Override
+    public boolean tryConsumeApplicationPersistedMsg(String clientId) {
+        if (!applicationPersistedMsgsRateLimitsConfiguration.isEnabled()) {
+            return true;
+        }
+        TbRateLimits rateLimits = applicationPersistedMsgClientLimits.computeIfAbsent(
+                clientId, id -> new TbRateLimits(applicationPersistedMsgsRateLimitsConfiguration.getClientConfig()));
+        return rateLimits.tryConsume();
+    }
+
+    @Override
+    public boolean isApplicationPersistedMsgsRateLimitEnabled() {
+        return applicationPersistedMsgsRateLimitsConfiguration.isEnabled();
     }
 
     @Override
--- a/application/src/main/java/org/thingsboard/mqtt/broker/service/limits/RateLimitServiceImpl.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/service/limits/RateLimitServiceImpl.java
@@ -186,7 +186,7 @@ public class RateLimitServiceImpl implements RateLimitService {
     }
 
     @Override
-    public boolean tryConsumeApplicationPersistedMsg(String clientId) {
+    public boolean tryConsumeApplicationPersistedMsgs(String clientId) {
         if (!applicationPersistedMsgsRateLimitsConfiguration.isEnabled()) {
             return true;
         }
--- a/application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/persistence/application/ApplicationPersistenceProcessorImpl.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/persistence/application/ApplicationPersistenceProcessorImpl.java
@@ -457,6 +457,7 @@ public class ApplicationPersistenceProcessorImpl implements ApplicationPersisten
 
         List<PersistedMsg> messagesToDeliver = buildMessagesToDeliver(pubRelMsgCtx, clientSessionCtx, persistedMsgCtx, messages, null);
         submitStrategy.init(messagesToDeliver);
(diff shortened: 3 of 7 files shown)


dmytro-landiak (commented):
## Review summary

Reviewed 7 changed files in [Outbound rate limiting for persistent APPLICATION clients](https://github.com/thingsboard/tbmq/pull/349). Left 7 comment(s) inline.

The feature is cleanly scoped and correctly opt-in (default off), reuses the existing `TbRateLimits`/bucket4j framework, and the no-data-loss reasoning holds up: tokens are consumed before delivery, offsets commit only after acks, and both the inactive-session and interrupt paths return without committing, so an un-throttled pack is simply re-polled. The shared-subscription rebalance caveat is already documented with a follow-up. Findings are mostly quality/consistency; the one correctness question is about message-expiry being evaluated before the throttle wait.

---

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/persistence/application/ApplicationPersistenceProcessorImpl.java
```diff
@@ -453,6 +457,7 @@ private ApplicationPubRelMsgCtx processMainPack(ApplicationPubRelMsgCtx pubRelMs
 
         List<PersistedMsg> messagesToDeliver = buildMessagesToDeliver(pubRelMsgCtx, clientSessionCtx, persistedMsgCtx, messages, null);
         submitStrategy.init(messagesToDeliver);
+        throttleDelivery(clientId, messagesToDeliver, () -> isClientSessionActive(sessionId, clientState));
```
dmytro-landiak, line 460: Message expiry is evaluated once in `buildPublishMessagesToDeliver` (against `System.currentTimeMillis()` at build time), but `throttleDelivery` can then hold the pack here for seconds — or much longer at low limits — before `deliverMessages` runs. A message that was still within its expiry window when the pack was built could cross the boundary during the throttle wait and then be delivered anyway. Is that acceptable for MQTT message-expiry semantics, or should expiry be re-checked after throttling / just before the send?

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/persistence/application/ApplicationPersistenceProcessorImpl.java
```diff
@@ -731,6 +737,46 @@ private void deliverMessages(ApplicationSubmitStrategy submitStrategy, ClientSes
         appMsgDeliveryStrategy.process(submitStrategy, clientSessionCtx);
     }
 
+    void throttleDelivery(String clientId, List<PersistedMsg> messagesToDeliver, BooleanSupplier isActive) {
+        if (!rateLimitService.isApplicationPersistedMsgsRateLimitEnabled()) {
+            return;
+        }
+        int remaining = countPublishMsgs(messagesToDeliver);
```
dmytro-landiak, line 744: `throttleDelivery` acquires tokens for the whole pack before `deliverMessages` runs, so the pack is still handed off as a single burst once enough tokens accrue — the loop paces token *acquisition*, not the actual sends. With e.g. `100:1,5000:60` and a large polled pack (main up to 200, shared up to 500), the client sees nothing for a few seconds and then receives the whole pack at once rather than a steady drip. The bucket does bound the long-run average, so this is a reasonable simplification — just want to confirm it's intentional given the PR describes the goal as replaying at a steady rate. Was interleaving the gate with per-message delivery considered?

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/persistence/application/ApplicationPersistenceProcessorImpl.java
```diff
@@ -731,6 +737,46 @@ private void deliverMessages(ApplicationSubmitStrategy submitStrategy, ClientSes
...
+            if (rateLimitService.tryConsumeApplicationPersistedMsgs(clientId)) {
+                remaining--;
+            } else {
+                if (!throttled) {
+                    throttled = true;
+                    if (isDebugEnabled) {
+                        log.debug("[{}] Outbound rate limit reached; pacing delivery of {} remaining message(s) in this pack", clientId, remaining);
+                    }
+                }
+                // Reuse the Kafka poll interval as the throttle back-off granularity (intentionally no separate tunable).
```
dmytro-landiak, line 759: Reusing `queue.application-persisted-msg.poll-interval` as the back-off granularity couples the throttle to a value that's really tuned for Kafka polling. If someone later changes the poll interval for Kafka reasons, the throttle back-off silently changes with it (and vice versa). The comment shows it's deliberate, which helps — just flagging that the coupling isn't visible from the call sites, and a small dedicated default might age better.

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/persistence/application/ApplicationPersistenceProcessorImpl.java
```diff
@@ -558,6 +563,7 @@ private ApplicationPubRelMsgCtx processSharedPack(ApplicationPubRelMsgCtx pubRel
 
         List<PersistedMsg> messagesToDeliver = buildMessagesToDeliver(pubRelMsgCtx, clientSessionCtx, persistedMsgCtx, messages, subscription);
         submitStrategy.init(messagesToDeliver);
+        throttleDelivery(clientId, messagesToDeliver, () -> isJobActive(job));
```
dmytro-landiak, line 566: The four new unit tests exercise `throttleDelivery` directly, but nothing verifies the wiring — that `processMainPack`/`processSharedPack` actually invoke it before the retry loop with the right `isActive` supplier. The existing `processMainPack` tests run with the limit mocked disabled (so the gate is a no-op), and the shared path has no throttling test at all. A test that drives these methods with the limit enabled and asserts `tryConsumeApplicationPersistedMsgs` is called would lock in the placement — it's easy to move this call to the wrong spot and not notice.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/RateLimitServiceImpl.java
```diff
@@ -175,6 +185,21 @@ public boolean isDevicePersistedMsgsLimitEnabled() {
         return devicePersistedMsgsRateLimitsConfiguration.isEnabled();
     }
 
+    @Override
+    public boolean tryConsumeApplicationPersistedMsgs(String clientId) {
+        if (!applicationPersistedMsgsRateLimitsConfiguration.isEnabled()) {
+            return true;
+        }
+        TbRateLimits rateLimits = applicationPersistedMsgClientLimits.computeIfAbsent(
```
dmytro-landiak, line 193: This is now the third copy of the per-client bucket idiom — `checkIncomingLimits` (line 77) and `checkOutgoingLimits` (line 93) both do the same `map.computeIfAbsent(clientId, id -> new TbRateLimits(config.getClientConfig())).tryConsume()`. Would it be worth extracting a small private helper like `tryConsume(ConcurrentMap<String, TbRateLimits> map, String clientConfig, String clientId)` and having the three call sites supply their own map/config? The surrounding logic differs (logging, the QoS-0 gate), so only the two-line core would move, but it keeps the bucket-creation contract in one place.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/RateLimitService.java
```diff
@@ -46,6 +46,13 @@ public interface RateLimitService {
 
     boolean isDevicePersistedMsgsLimitEnabled();
 
+    /**
+     * Non-blocking token consume for outbound persisted messages to an APPLICATION client (throttling, not dropping)
+     */
+    boolean tryConsumeApplicationPersistedMsgs(String clientId);
+
+    boolean isApplicationPersistedMsgsRateLimitEnabled();
```
dmytro-landiak, line 54: The sibling flags in this interface are `isDevicePersistedMsgsLimitEnabled()` and `isTotalMsgsLimitEnabled()`, but this one has an extra `Rate`: `isApplicationPersistedMsgsRateLimitEnabled()`. Minor, but for a set of methods a caller finds via autocomplete the odd-one-out is a small papercut — `isApplicationPersistedMsgsLimitEnabled()` would line up with the others.

application/src/test/java/org/thingsboard/mqtt/broker/service/mqtt/persistence/application/ApplicationPersistenceProcessorImplTest.java
```diff
@@ -575,6 +580,58 @@ void processMainPack_retryAll_mustObtainAckStrategyOncePerPack_notOncePerRetry()
         verify(acknowledgeStrategyFactory, times(1)).newInstance("appClient");
     }
 
+    @Test
+    public void givenRateLimitDisabled_whenThrottleDelivery_thenNoTokensConsumed() {
```
dmytro-landiak, line 584: These four throttle tests use `public` visibility and given/when/then names, while the rest of this JUnit 5 file uses package-private methods and `method_scenario_result` names (e.g. `processMainPack_retryAll_whenClientNeverAcks_...`). Matching the surrounding convention keeps the file consistent for the next reader.

PR #354, opened 2026-08-07
feat(limits): always-on cluster-wide total throughput quota with packet-accurate charging

dmytro-landiak:
## Pull Request description

Reworks `mqtt.rate-limits.total` from the legacy four-site implementation into an always-on, cluster-wide **total throughput quota** with packet-accurate charging.

### What changed

- **New `ThroughputQuotaService`** (`service.limits`): a node-local token lease fronted by asynchronous single-flight block draws from the shared Redis bucket. Callers never touch Redis on the hot path; while a draw is pending, demand rides bounded credit (down to `-block-size`); a zero-grant draw opens a 50 ms local backoff; unused tokens return to the bucket on a periodic tick (`lease-return-ms`).
- **Fail-open Redis semantics:** if Redis is unavailable — including while a failing draw is still in flight — traffic passes unmetered rather than being refused. Degradations are observable via the new `throughputQuotaDegraded{cause=redis}` meter and a rate-limited WARN when the quota starts clamping.
- **The billed unit is one PUBLISH message processed** — a wire PUBLISH packet, or a message routed to an integration — charged at six sites:
  - ingress (`MqttSessionHandler`) — refusal: MQTT 5 gets PUBACK/PUBREC reason `0x97 QUOTA_EXCEEDED` and stays connected; MQTT 3.x is disconnected; a refused packet no longer retains the payload buffer;
  - basic downlink (`BasicDownLinkProcessorImpl`) — refused deliveries are dropped and reported in `droppedMsgs`;
  - device-persisted deliveries (`PersistedDeviceActorMessageProcessor`) — one choke point for live delivery, reconnect replay, and shared-subscription catch-up; refusals settle terminally (message removed, `droppedMsgs` reported, no counter/in-flight mutation); PUBREL redeliveries ride free;
  - application pack rounds (`ApplicationPersistenceProcessorImpl`) — refused PUBLISH tail is excluded before pack-ctx creation, so latch/stats stay exact and offsets commit normally (no stall, no redelivery); PUBRELs ride free;
  - integration fan-out (`MsgPersistenceManagerImpl`) — the single enqueue point to the integration downlink topic; the granted prefix is sent, the refused remainder settles on the pack callback and is reported in `droppedMsgs`;
  - retained messages (`MqttSubscribeHandler`) — the retained list is truncated to the granted prefix; deliberately **no** `droppedMsgs` report, since the message stays in the retain store and re-sends on the next matching SUBSCRIBE.
- **Deleted legacy machinery:** `RateLimitBatchProcessor`/`RateLimitBatchProcessorImpl` (+ test), the dispatcher fan-out charge, the consumer poll charge, and `RateLimitService.tryConsumeTotalMsgs`/`isTotalMsgsLimitEnabled`. Per-client rate limits are untouched.
- **Configuration** (`mqtt.rate-limits.total`): new `block-size` (`MQTT_TOTAL_RATE_LIMITS_BLOCK_SIZE`, default `0` = auto-derive `max(1, sustainedRate / 10)`) and `lease-return-ms` (`MQTT_TOTAL_RATE_LIMITS_LEASE_RETURN_MS`, default `1000`, `0` disables). `enabled`/`config` and their env vars are preserved.

### Behavior changes to call out

- A publish delivered to N subscribers costs N+1 packets; a shared-subscription group is charged once per message; persisted-session replay and QoS 1/2 retransmissions are counted; messages routed to integration subscribers count as outgoing deliveries (one per matched integration, as before this rework).
- Enforcement is always-on when enabled: an accepted QoS 1/2 publish may still have its deliveries dropped later under overuse, and a device reconnecting under an exhausted quota can have the refused part of its backlog terminally removed — capacity should be sized as a burst allowance.
- **Config-breaking:** `MQTT_RATE_LIMITS_THREADS_COUNT`, `MQTT_RATE_LIMITS_BATCH_SIZE`, and `MQTT_RATE_LIMITS_PERIOD_MS` are removed together with the batch-processing implementation they configured (only affects deployments that set them).

### Documentation notes

The rate-limits documentation needs: the packet-unit semantics above (N+1 fan-out, shared group once, replay/retransmissions counted, integration deliveries counted), the two new parameters, the three removed parameters, the refusal behavior per MQTT version, and the fail-open + `throughputQuotaDegraded` observability note.

## General checklist

- [ ] [Labels](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels#about-labels) that classify your pull request have been added.
- [ ] The [milestone](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/about-milestones) is specified and corresponds to fix version.
- [ ] Description references specific [issue](https://github.com/thingsboard/tbmq/issues).
- [ ] Changes are backward compatible or upgrade script is provided. *(Config-breaking removals are documented above; no upgrade script — the removed parameters have no replacement to migrate to.)*

*(Not applicable — no UI changes.)*

- [ ] Screenshots with affected component(s) are added. The best option is to provide 2 screens: before and after changes;
- [ ] If you change the widget or other API, ensure it is backward-compatible or upgrade script is present.

## Back-End feature checklist

- [x] If new dependency was added: the dependency tree is checked for conflicts. *(No new dependencies.)*


application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
```diff
@@ -0,0 +1,215 @@
...
+        // once a draw confirmed the bucket dry, refuse locally until the backoff elapses
+        long creditFloor = now - dryUntilNanos < 0 ? 0L : -(long) blockSize;
+        while (true) {
+            long current = localTokens.get();
+            long available = current - creditFloor;
+            if (available <= 0) {
+                scheduleDrawIfNeeded(n);
+                // the fail-open window above lapses while the next draw is still hanging on an unreachable
+                // Redis, so without this the quota would turn fail-closed for that draw's whole duration
+                return redisDegraded ? n : 0;
```
dmytro-landiak: The fail-open contract has a gap on the healthy→down transition: `redisDegraded` only latches after a draw *completes* exceptionally, so when Redis dies mid-operation the first failing draw hangs with `redisDegraded == false` for its whole socket timeout. Once the one block of bounded credit drains (milliseconds under load), every call lands here and returns 0 — a fail-closed window of potentially seconds per outage. That contradicts the stated "including while a failing draw is still in flight" semantics, which only hold once a *previous* draw has already failed and latched the flag. The field comment above accepts exactly this for startup ("one block, deliberately accepted") — was extending that acceptance to any mid-life outage deliberate? If the window matters, one option is to also treat "a draw has been in flight longer than X" as degraded.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    private void scheduleDrawIfNeeded(long shortfall) {
+        if (System.nanoTime() - dryUntilNanos < 0) {
+            return; // bucket known dry: no Redis traffic during the backoff window
+        }
+        if (!drawInFlight.compareAndSet(false, true)) {
+            return; // single-flight: one draw at a time per node
+        }
+        long drawSize = Math.max(blockSize, shortfall);
```
dmytro-landiak, line 165: `Math.max(blockSize, shortfall)` sizes the draw to demand that was already terminally refused — the drawn tokens only ever serve *future* traffic. With the default 1 s lease return that's self-correcting, but with `lease-return-ms: 0` (documented as a supported way to disable the tick) a single large refused bulk charge — e.g. a wildcard SUBSCRIBE matching thousands of retained messages — strands up to `shortfall` tokens on this node permanently, starving the rest of the cluster. Worth capping the draw at `blockSize` when the return tick is disabled, or at least calling the interaction out in the yml docs?

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
```diff
@@ -0,0 +1,215 @@
...
+    static final long DRY_BACKOFF_NANOS = TimeUnit.MILLISECONDS.toNanos(50);
+    static final long FAIL_OPEN_NANOS = TimeUnit.SECONDS.toNanos(1);
+    static final long CLAMP_WARN_INTERVAL_NANOS = TimeUnit.SECONDS.toNanos(60);
+
+    private final TotalMsgsRateLimitsConfiguration totalMsgsRateLimitsConfiguration;
+    private final ThroughputLimitProvider throughputLimitProvider;
+    private final RateLimitCacheService rateLimitCacheService;
+    private final StatsManager statsManager;
+
+    @Value("${mqtt.rate-limits.total.block-size:0}")
```
dmytro-landiak, line 50: The two new knobs under `mqtt.rate-limits.total` are `@Value`-injected here, while their siblings `enabled` and `config` bind through `TotalMsgsRateLimitsConfiguration` (`@ConfigurationProperties` on the same prefix). The next person adding a knob to this prefix will look in the configuration class and miss these. Any reason not to add `blockSize`/`leaseReturnMs` as fields on `TotalMsgsRateLimitsConfiguration`, which this service already receives?

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
```diff
@@ -0,0 +1,215 @@
...
+        failOpenUntilNanos = now;
+        lastClampWarnNanos = now - CLAMP_WARN_INTERVAL_NANOS - 1; // the very first clamp warns immediately
+        blockSize = configuredBlockSize > 0 ? configuredBlockSize : deriveDefaultBlockSize();
+        stats = statsManager.getThroughputQuotaStats();
+        if (drawExecutor == null) {
+            drawExecutor = ThingsBoardExecutors.initSingleExecutorService("throughput-quota-draw");
+        }
+        if (leaseReturnScheduler == null && leaseReturnMs > 0) {
+            leaseReturnScheduler = ThingsBoardExecutors.newSingleScheduledThreadPool("throughput-quota-lease-return");
+            leaseReturnScheduler.scheduleAtFixedRate(this::returnUnusedTokens, leaseReturnMs, leaseReturnMs, TimeUnit.MILLISECONDS);
```
dmytro-landiak, line 90: The lease-return scheduling is never exercised by any test: the unit tests call `returnUnusedTokens()` by hand (`leaseReturnMs` is 0 there, so `init()` never creates the scheduler), and all six integration tests pin `lease-return-ms=600000`, which disables the tick for the test's duration. So one of the two new config knobs has no end-to-end proof that idle nodes actually give budget back, and the documented `lease-return-ms: 0` disable path is only hit as an accidental field default. An integration case with a short lease-return period (idle node returns its block, a second consumer gets it) would close both gaps.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
```diff
@@ -0,0 +1,215 @@
...
+        if (current <= 0) {
+            return; // never return a credit deficit
+        }
+        if (!localTokens.compareAndSet(current, 0)) {
+            return; // raced with consumption; the next tick handles it
+        }
+        try {
+            rateLimitCacheService.returnTotalMsgs(current);
+        } catch (Exception e) {
+            localTokens.addAndGet(current); // keep the budget locally rather than losing it
```
dmytro-landiak, line 219: This compensation branch — Redis throws during the lease return and the tokens are restored locally — has no test. The happy path and the credit-deficit guard are covered in `ThroughputQuotaServiceImplTest`, but nothing stubs `returnTotalMsgs` to throw and asserts the balance survives (e.g. a subsequent `tryConsumeOutgoing` still grants from the restored budget). Given how much care the draw-side failure handling gets in tests, this path deserves one too.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
```diff
@@ -0,0 +1,215 @@
...
+            log.debug("Failed to return unused total throughput quota tokens to Redis", e);
+        }
+    }
+
+    int deriveDefaultBlockSize() {
+        return (int) Math.min(Integer.MAX_VALUE, Math.max(1, throughputLimitProvider.getSustainedRatePerSec() / 10));
```
dmytro-landiak, line 225: The `/ 10` is a magic number that's also independently documented in the YAML comment ("derives max(1, sustainedRate / 10)") — if one changes, the other silently drifts. A named constant (e.g. `DEFAULT_DRAWS_PER_SECOND = 10`) alongside the other tuning constants at the top of the class would name the intent: roughly ten Redis draws per second at sustained rate.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaService.java
```diff
@@ -0,0 +1,42 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.thingsboard.mqtt.broker.service.limits;
+
+/**
+ * Cluster-wide total throughput quota — the single policy owner for the billed unit: MQTT PUBLISH
+ * packets processed, incoming and outgoing combined; messages routed to integration subscribers
+ * count as outgoing (spec: docs/superpowers/specs/2026-08-06-total-throughput-quota-design.md).
```
dmytro-landiak, line 21: This javadoc cites `docs/superpowers/specs/2026-08-06-total-throughput-quota-design.md`, but the whole `/docs/` folder is gitignored (widened in this very PR), so the reference is a dead link for anyone cloning the repo — same for the "spec §4.3" pointer in `ThroughputLimitProvider`'s javadoc. Since these interfaces are the policy owners, could the load-bearing rules (billed unit, charge sites, fail-open contract) be stated inline instead of pointing at a document that will never exist in the repository?

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/DefaultThroughputLimitProvider.java
```diff
@@ -0,0 +1,40 @@
...
+@Service
+@RequiredArgsConstructor
+public class DefaultThroughputLimitProvider implements ThroughputLimitProvider {
+
+    private final TotalMsgsRateLimitsConfiguration totalMsgsRateLimitsConfiguration;
+
+    @Override
+    public long getSustainedRatePerSec() {
+        long minRatePerSec = Long.MAX_VALUE;
+        for (String limitSrc : totalMsgsRateLimitsConfiguration.getConfig().split(BrokerConstants.COMMA)) {
```
dmytro-landiak, line 32: This re-parses the `capacity:seconds` config string, duplicating the parse in `AbstractMsgsRateLimitsConfiguration.getBucketConfiguration()` — and without its `parts.length != 2` validation, so the two parses can drift. Could the configuration class expose the parsed bandwidths (or a `getMinSustainedRatePerSec()` helper) so the format is parsed in exactly one place? Alternatively derive the rate from the `totalMsgsBucketConfiguration` bean's `Bandwidth` objects, which already carry capacity and refill period. Also, the `Math.max(1, ...)` clamp below — the branch that matters for sub-1-msg/s bandwidths like the `10:600` configs every new integration test uses — is untested, and this class's only direct test (the multi-bandwidth min case) lives inside `ThroughputQuotaServiceImplTest`; a small dedicated test would document the derivation rules where a reader will look for them.

application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/handlers/MqttSubscribeHandler.java
```diff
@@ -213,24 +213,18 @@ private void processRetainedMessages(ClientSessionCtx ctx,
...
     List<RetainedMsg> applyRateLimits(List<RetainedMsg> retainedMsgList) {
-        if (!rateLimitService.isTotalMsgsLimitEnabled()) {
-            return retainedMsgList;
-        }
-
         int totalMsgCount = retainedMsgList.size();
-        int availableTokens = (int) rateLimitService.tryConsumeTotalMsgs(totalMsgCount);
-
-        if (availableTokens >= totalMsgCount) {
+        int granted = throughputQuotaService.tryConsumeOutgoing(totalMsgCount);
```
dmytro-landiak, line 217: The method is still named `applyRateLimits`, but after this change it applies the throughput quota — its own log lines say "total throughput quota", and the class separately holds a real `RateLimitService`, so the name now actively misleads. Renaming to `applyThroughputQuota` would also match the sibling site `ApplicationPersistenceProcessorImpl.applyThroughputQuota`; the test names (`givenQuotaExhausted_whenApplyRateLimits_...`) carry the same staleness.

application/src/main/java/org/thingsboard/mqtt/broker/service/processing/downlink/basic/BasicDownLinkProcessorImpl.java
```diff
@@ -49,6 +51,10 @@ public void process(String clientId, PublishMsgProto msg) {
             dropMessage();
             return;
         }
+        if (throughputQuotaService.tryConsumeOutgoing(1) == 0) {
```
dmytro-landiak, line 54: Three call sites (both `process` overloads here plus `PersistedDeviceActorMessageProcessor.processPublishMsg`) charge a single packet via `tryConsumeOutgoing(1) == 0`. A boolean single-packet convenience on the interface — mirroring `tryConsumeIncoming()` — would make these read as a grant/refuse decision instead of an integer comparison and keep the API symmetric for the common single-message case.

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/persistence/MsgPersistenceManagerImpl.java
```diff
@@ -153,6 +153,29 @@ void processDeviceSubscriptionsWithRateLimits(List<Subscription> deviceSubscript
         }
     }
 
+    void processIntegrationSubscriptionsWithThroughputQuota(List<Subscription> integrationSubscriptions,
```
dmytro-landiak, line 156: This is structurally the same algorithm as `processDeviceSubscriptionsWithRateLimits` directly above: bulk try-consume, full-grant fast path, report dropped, send the granted prefix, settle the remainder through the callback. The two diverge only in token source and settle mechanics (`onBatchSuccess(dropped)` here vs a per-message `onSuccess()` loop there). Worth either extracting the shared "deliver granted prefix, settle rest" template with the send function and token draw injected, or at least aligning the settle idiom so the two methods read identically — right now a reader has to diff them to learn the difference is cosmetic.

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/persistence/MsgPersistenceManagerImpl.java
```diff
@@ -153,6 +153,29 @@ void processDeviceSubscriptionsWithRateLimits(List<Subscription> deviceSubscript
...
+        }
+
+        int dropped = totalCount - granted;
+        log.trace("Total throughput quota exceeded. Dropping {} integration messages", dropped);
+        tbMessageStatsReportClient.reportDroppedMsgs(dropped);
+
+        for (int i = 0; i < granted; i++) {
+            sendIntegrationMsg(integrationSubscriptions.get(i), publishMsgWithId, callbackWrapper);
+        }
+        callbackWrapper.onBatchSuccess(dropped);
```
dmytro-landiak, line 176: Of the six charge sites, this integration fan-out one is the only site verified solely with mocks — the other five each got a dedicated end-to-end `IntegrationTestCase`. The partial-grant contract here (first-N-by-list-order prefix delivered, remainder settled so the pack callback completes) is exactly the kind of behavior the other tests prove through the real flow. If a real integration executor can't run in this suite, is `msa/black-box-tests` the right home for it? Worth a note in the PR either way so the gap is a decision, not an accident.

.gitignore
```diff
@@ -36,6 +36,6 @@ rebuild-docker.sh
 */.run/**
 .run/**
 .run
-/docs/superpowers/
+/docs/
```
dmytro-landiak, line 39: Widening `/docs/superpowers/` to `/docs/` is repo-hygiene work unrelated to the throughput quota, and it ships to every contributor — any future `docs/` directory in the repo is silently untracked. If the intent is a local-only docs folder, `.git/info/exclude` achieves that without affecting others. Intentional to put it in the shared `.gitignore` (and in this PR), especially given the new javadoc references a spec under `docs/` that this line permanently untracks?

dmytro-landiak (commented):
## Review summary

Reviewed 47 changed files in [feat(limits): always-on cluster-wide total throughput quota with packet-accurate charging](https://github.com/thingsboard/tbmq/pull/354). Left 18 comment(s) inline.

The core lease/draw/fail-open machinery in `ThroughputQuotaServiceImpl` is carefully built and unusually well tested (single-flight release on executor failure, degraded-flag semantics, credit accounting), and the six charge sites settle refusals consistently — I verified there is no payload-buffer leak on ingress refusal, callback accounting on the integration fan-out is exact, the device-actor settlement mirrors the ack path (including the shared-subscription key mapping), PUBRELs ride free on both persisted paths, and there is no double-charging across sites. The inline comments are mostly design questions and test-coverage/maintainability points; the one behavioral gap worth a close look is the fail-closed window on a healthy→down Redis transition while the first failing draw is still in flight.

---

application/src/main/resources/thingsboard-mqtt-broker.yml
```diff
@@ -1008,18 +1008,29 @@ mqtt:
...
-    # Period to wait before processing a batch of messages for total rate limits (in ms).
-    period-ms: "${MQTT_RATE_LIMITS_PERIOD_MS:50}"
     total:
-      # Enables or disables total incoming and outgoing messages rate limits for the broker (per whole cluster).
+      # Enables or disables the cluster-wide total throughput quota. The quota counts MQTT PUBLISH packets processed by
+      # the broker — both incoming (from publishers) and outgoing (to subscribers, including messages routed to
+      # integrations) — against one budget shared by the whole cluster. A publish delivered to N subscribers costs N+1
+      # packets; persisted-session replay and QoS 1/2 retransmissions are counted; a shared subscription group is
+      # counted once per message. On overuse, incoming publishes are refused (MQTT 5: PUBACK/PUBREC with reason code
+      # 0x97 QUOTA_EXCEEDED; MQTT 3.x: disconnect) and outgoing deliveries are dropped and reported in the droppedMsgs
```
dmytro-landiak, line 1017: Tiny doc nit: for MQTT 5 QoS 0 there is no PUBACK/PUBREC to carry 0x97, so a refused QoS 0 publish is silently dropped while the client stays connected — worth a clause here, since this comment is the behavioral contract operators will read.

application/src/test/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImplTest.java
```diff
@@ -0,0 +1,319 @@
...
+    public void givenDryBucket_whenBackoffActive_thenRefusesLocallyAndRecoversAfterBackoff() throws InterruptedException {
+        when(rateLimitCacheService.tryConsumeTotalMsgs(anyLong())).thenReturn(10L, 0L, 5L);
+        service.drawExecutor = MoreExecutors.newDirectExecutorService();
+        service.init(); // draw#1: local = 10
+
+        assertEquals(10, service.tryConsumeOutgoing(10)); // exhausts local -> draw#2 returns 0 -> dry backoff opens
+        assertFalse("dry backoff must refuse without credit", service.tryConsumeIncoming());
+        verify(rateLimitCacheService, times(2)).tryConsumeTotalMsgs(anyLong()); // dry window: refusals must not reach Redis
+
+        Thread.sleep(60); // DRY_BACKOFF_NANOS is 50 ms
```
dmytro-landiak, line 198: These sleeps hardcode the current values of `DRY_BACKOFF_NANOS` and `FAIL_OPEN_NANOS` (also the 1100 ms sleep further down) even though the constants are package-visible from this test. Deriving the sleep as `TimeUnit.NANOSECONDS.toMillis(DRY_BACKOFF_NANOS) + margin` would keep the tests correct if the windows are ever tuned, instead of failing mysteriously — the comments already show the coupling exists.

application/src/test/java/org/thingsboard/mqtt/broker/service/testing/integration/TotalThroughputQuotaAppPersistedIntegrationTestCase.java
```diff
@@ -0,0 +1,169 @@
...
+        // reconnecting consumer assigns its partition and polls immediately. Reconnecting too early can therefore
+        // split the 6 messages across two polls, and since applyThroughputQuota charges once PER PACK the second pack
+        // re-arms the block credit and all 6 get delivered, breaking the `< 6` assertion. At 1 token per 60 s this
+        // settle adds ~0.017 tokens, so it leaves the ingress ledger untouched.
+        Thread.sleep(1000);
+
+        persistedClient.connect(appOptions);                // pack replay: one bulk charge grants 2 local + 2 credit
+        Thread.sleep(5000);
+        int deliveredOnReplay = received.get();
+        assertTrue("pack must make progress (no stall)", deliveredOnReplay >= 4);
```
dmytro-landiak, line 161: The ledger arithmetic (excellently documented) assumes every async Redis draw lands inside the 50 ms publish pacing; on a loaded CI worker a slow draw makes an ingress charge ride credit, the local balance at reconnect drops below 2, and the `>= 4` lower bound fails even though the feature is behaving correctly. The device-persisted sibling deliberately asserts a 2..4 band for exactly this reason — should this lower bound be similarly relaxed? Separately, the fixed `Thread.sleep(5000)`/`Thread.sleep(3000)` waits cost ~10 s of wall time per run; Awaitility (already used in this class) polling for received-count stability would be both faster and less timing-sensitive.

application/src/test/java/org/thingsboard/mqtt/broker/service/testing/integration/TotalThroughputQuotaAppPersistedIntegrationTestCase.java
```diff
@@ -0,0 +1,169 @@
...
+        deviceCredentials = credentialsService.saveCredentials(
+                TestUtils.createDeviceClientCredentials(PUBLISHING_CLIENT, DEV_USERNAME)
+        );
+        enableBasicProvider();
+    }
+
+    @After
+    public void clear() throws Exception {
+        log.warn("After test finish...");
+        if (persistedClient.isConnected()) {
```
dmytro-landiak, line 117: If the test (or `beforeTest`) fails before `persistedClient` is assigned, this NPEs and masks the original failure — the device-persisted sibling guards its `@After` with `if (persistedClient != null)`. More broadly, this teardown (reconnect clean-session to wipe the persistent session, including the 30-second Awaitility poll running inside a log argument) is copied verbatim from `AppPersistedMessagesIntegrationTestCase` — roughly the tenth copy of this cleanup across the suite. Extracting it (and `getConnectOptions`) into a shared base/helper would stop the spread.

application/src/test/java/org/thingsboard/mqtt/broker/service/testing/integration/TotalThroughputQuotaRefusalIntegrationTestCase.java
```diff
@@ -0,0 +1,121 @@
...
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_CLASS)
+@ContextConfiguration(classes = TotalThroughputQuotaRefusalIntegrationTestCase.class, loader = SpringBootContextLoader.class)
+@TestPropertySource(properties = {
+        "mqtt.rate-limits.total.enabled=true",
+        "mqtt.rate-limits.total.config=5:600",
+        "mqtt.rate-limits.total.block-size=1",
+        "mqtt.rate-limits.total.lease-return-ms=600000"
+})
+@DaoSqlTest
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
```
dmytro-landiak, line 49: `b_` leans on `a_` having drained the shared bucket ("the bucket was exhausted by test a_, so refusals MUST appear") — alphabetical method-order coupling through shared Redis state. It happens to still pass standalone (10 publishes vs capacity 5), but a rename or a future third method silently changes what each test proves. Making each method self-sufficient (or folding both protocols into one method with an explicit sequence) would remove the hidden ordering contract.

dmytro-landiak:
Superseded by #356, which carries the same feature with a smaller fix for the transient node-local shortfall: a right-sized draw on the calling thread instead of deferring and retrying the refused remainder. 60 lines across 4 files instead of 269 across 7, no `QuotaGrant` type, no retry message, no pack-commit gating — and it also closes the retained-truncation case this branch left open.

Both approaches were measured and both eliminate the loss entirely. The tradeoff, along with the latency cost that wait-for-draw pays under multi-node bucket contention, is recorded in #356's "Alternatives considered" section.

Closing rather than force-pushing so this implementation stays readable for comparison. Deleting the branch — the commits remain accessible through this PR.

PR #356, opened 2026-08-12
feat(limits): cluster-wide total throughput quota charged before persistence

dmytro-landiak:
## Pull Request description

Adds the cluster-wide **total throughput quota**: one budget, shared by every node, counted in MQTT PUBLISH packets. The quota is opt-in and disabled by default (`mqtt.rate-limits.total.enabled=false`), so nothing changes for existing deployments unless it is turned on.

**Charging happens at admission, before durability.** A publish costs 1 packet for the incoming message plus 1 for every persistent subscriber, shared subscription group and integration it is routed to, all charged on the publish path in `MsgPersistenceManagerImpl`. Once a message has been charged and stored, delivering it costs nothing more.

What that buys, and why it is the model:

- **A backlog is never destroyed.** A persistent session that reconnects after being offline receives everything that was stored for it, however large, and QoS 1/2 retransmissions are free. Charging at delivery instead meant an exhausted quota discarded already-durable messages — and, for APPLICATION clients, discarded them *after* the pack had committed their offsets, so they never came back. A refusal at admission sheds work the broker never promised to do; a refusal at delivery destroys work it already acknowledged.
- **The bill is deterministic.** Cost is `1 + fanout` computed at publish time, independent of whether subscribers happened to be online.
- **Over-quota drops move from delivery time to publish time.** When the remaining budget cannot cover a fan-out, the copies beyond the cut are never stored and are reported in the `droppedMsgs` statistic. Truncation drops the tail of the subscription list.

Enforcement is node-local and non-blocking on the hot paths: each node draws tokens from the shared Redis (Valkey) bucket in blocks, charges against that pool atomically, and returns what it did not use on a background tick. The persistent fan-out is the one charge that may draw synchronously — it runs on the publish-dispatcher Kafka consumer thread, and sizing the draw to real demand is what keeps a wide fan-out bounded by the cluster budget rather than by the node's lease. If Redis is unreachable the quota **fails open**: traffic passes unmetered and the `throughputQuotaDegraded` metric is raised.

Behaviour on overuse:

| Packet | Result |
|---|---|
| Incoming PUBLISH | refused — MQTT 5: PUBACK/PUBREC with `0x97 QUOTA_EXCEEDED`; MQTT 3.x: disconnect |
| Persistent subscriber / shared group / integration copy | never stored, reported in `droppedMsgs` |
| Non-persistent (QoS 0) delivery | dropped, reported in `droppedMsgs` — delivery time *is* admission time here, nothing was stored |
| Retained message on SUBSCRIBE | truncated, retain store left intact so the next subscriber still gets it |

### Configuration

New parameters under the existing `mqtt.rate-limits.total` block in `thingsboard-mqtt-broker.yml`: `config` (bucket definition, default `50000:60`), `block-size` (tokens drawn per Redis round trip, `0` derives `sustainedRate / 10`) and `lease-return-ms` (unused-token return tick). `enabled` keeps its meaning and its `false` default.

**The bucket's capacity is also the largest amount a single publish can charge**, because a publish pays for its whole fan-out in one charge. The default is therefore a *single* bandwidth: `50000:60` sustains ~833 msgs/s while allowing a 50000-subscriber fan-out. Configuring more than one bandwidth lowers that per-publish ceiling to the smallest capacity among them, because bucket4j requires every bandwidth to hold the tokens for a single charge — `1000:1,50000:60` would sustain the same ~833/s but truncate any fan-out above 1000, even on an idle cluster. An operator who wants a smaller burst should shorten the period at the same ratio rather than add a bandwidth (`5000:6` is the same ~833/s with a tenth of the burst), with their widest fan-out as the floor on capacity. All of this is documented in the yml, and `TotalMsgsRateLimitsBucketShapeTest` pins both halves — that the single-bandwidth default really does admit a wide fan-out, and that the multi-bandwidth form really does cap it — so the guidance cannot drift from the behaviour.

### Release notes

Upgrade-facing changes in this PR, for the release notes:

- **`mqtt.rate-limits.total` has new semantics.** The total rate limit is now a cluster-wide throughput quota charged at **admission** rather than at delivery. Over-quota drops therefore happen when a publish is processed, not when it is delivered, and a stored backlog is always delivered in full regardless of quota state.
- **Three configuration parameters are removed**, along with the batch processor they configured: `MQTT_RATE_LIMITS_THREADS_COUNT`, `MQTT_RATE_LIMITS_BATCH_SIZE` and `MQTT_RATE_LIMITS_PERIOD_MS`. Deployments that set them will start normally, but the values are ignored.
- **The `MQTT_TOTAL_RATE_LIMITS_CONFIG` default changed from `1000:1,50000:60` to `50000:60`.** If you run with `MQTT_TOTAL_RATE_LIMITS_ENABLED=true` and never set the config explicitly, the per-second ceiling is gone: the sustained rate is unchanged at ~833 msgs/s, but a single second can now admit up to 50000 packets. This is deliberate — a publish charges for its entire fan-out at once, so any tighter bandwidth would also cap fan-out width at its own capacity. To keep a smaller burst, shorten the period at the same ratio (for example `5000:6`) rather than adding a second bandwidth.
- **Two new parameters**, both optional: `MQTT_TOTAL_RATE_LIMITS_BLOCK_SIZE` and `MQTT_TOTAL_RATE_LIMITS_LEASE_RETURN_MS`.
- **New metric** `throughputQuotaDegraded{cause=redis}`, which is the only signal that enforcement has silently stopped because Redis is unreachable.

### Documentation

The `mqtt.rate-limits.total` yml block documents the charging model, the overuse behaviour, the fail-open behaviour and the capacity sizing rule. The documentation site needs a section on the quota: what a packet costs, that stored messages are delivered unmetered, the `0x97` refusal, the `droppedMsgs` / `throughputQuotaDegraded` metrics, and how to size the bucket for a given fan-out and burst.

## General checklist

No front-end changes.

## Back-End feature checklist

Unit coverage for the quota service (draw sizing, bounded credit, dry backoff, fail-open, lease return), for the fan-out charge and truncation in `MsgPersistenceManagerImpl` (including that every refused subscription still settles its callback, so a truncated publish is still acknowledged and committed), for the bucket shape, and for the config parser — the sustained rate and the bucket are now derived from one shared bandwidth parser, so a malformed config fails identically down both paths.

Eight integration cases cover the ingress `0x97` refusal, retained truncation, fan-out truncation at persist, once-per-group shared-subscription charging, unmetered replay of an APPLICATION and a DEVICE backlog with the shared bucket held empty throughout, and both routes into the non-persistent charge site — a clean-session fan-out and a QoS 0 publish to persistent subscribers, each asserting the accounting identity `delivered + droppedMsgs == subscribers`. No new dependencies.


application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/handlers/MqttSubscribeHandler.java
```diff
@@ -213,24 +213,22 @@ private void processRetainedMessages(ClientSessionCtx ctx,
...
-            log.debug("No available tokens left for total msgs bucket during retained msg processing. Skipping {} messages", totalMsgCount);
+        if (granted <= 0) {
+            log.debug("Total throughput quota exhausted during retained msg processing. Skipping {} messages", totalMsgCount);
             return Collections.emptyList();
         }
-
-        log.debug("Hitting total messages rate limits on retained msg processing. Skipping {} messages", totalMsgCount - availableTokens);
-        return retainedMsgList.subList(0, availableTokens);
+        log.debug("Hitting total throughput quota on retained msg processing. Skipping {} messages", totalMsgCount - granted);
+        // deliberately NO droppedMsgs report: the message stays in the retain store and re-sends on the next matching SUBSCRIBE
```
dmytro-landiak: The "re-sends on the next matching SUBSCRIBE" part only holds for `retainHandling = SEND_AT_SUBSCRIBE`. For `SEND_AT_SUBSCRIBE_IF_NOT_YET_EXISTS` (the MQTT 5 value 1), `needSendRetainedForTopicSubscription` evaluates `ts -> !currentSubscriptions.contains(ts)` (`SubscriptionOptions.java:99`), so once the subscription exists a repeat SUBSCRIBE will not resend — the truncated retained message is gone for that client for good, and because this path deliberately skips `reportDroppedMsgs` there is no metric for it either.

So for that one option the quota produces a silent, permanent loss. Two options that would both close it: report `droppedMsgs` for the truncated tail (accepting the slight double-meaning of the counter), or treat retained-on-subscribe as all-or-nothing when the subscription is `SEND_AT_SUBSCRIBE_IF_NOT_YET_EXISTS`. At minimum the comment should say which retain-handling values it is true for.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
```diff
@@ -0,0 +1,334 @@
...
+        if (current <= 0) {
+            return; // never return a credit deficit
+        }
+        if (!localTokens.compareAndSet(current, 0)) {
+            return; // raced with consumption; the next tick handles it
+        }
+        try {
+            rateLimitCacheService.returnTotalMsgs(current);
+        } catch (Exception e) {
+            localTokens.addAndGet(current); // keep the budget locally rather than losing it
```
dmytro-landiak: `returnTotalMsgs` is a non-idempotent remote mutation (`bucket.addTokens`), so this restore is only correct when the call definitely did not land. On a Jedis socket timeout where Redis actually applied the `addTokens`, the shared bucket gets the tokens *and* the node keeps them locally — the cluster budget inflates by up to one lease per failed tick, silently and permanently (nothing ever reconciles it back).

It is bounded and rare, so it may well be an acceptable trade-off, but as written the code reads as if the restore is unconditionally safe. Would it be worth dropping the tokens instead on a timeout (conservative: the cluster under-grants by one lease until the next refill) or at least noting the over-grant in the comment so the next reader knows it is a deliberate choice?

application/src/main/resources/thingsboard-mqtt-broker.yml
```diff
@@ -1008,20 +1008,41 @@ mqtt:
         # Number of messages buffered in the channel before the flush is made. Used when `MQTT_APP_MSG_WRITE_AND_FLUSH` = false.
         buffered-msg-count: "${MQTT_APP_BUFFERED_MSG_COUNT:10}"
   rate-limits:
-    # Number of parallel threads dedicated to processing total rate limit checks for incoming messages.
-    threads-count: "${MQTT_RATE_LIMITS_THREADS_COUNT:1}"
```
dmytro-landiak, line 1012: Dropping `threads-count` / `batch-size` / `period-ms` is right — the batch processor is gone — but Spring binds these leniently, so any deployment that sets `MQTT_RATE_LIMITS_THREADS_COUNT`, `MQTT_RATE_LIMITS_BATCH_SIZE` or `MQTT_RATE_LIMITS_PERIOD_MS` will keep starting cleanly with those values now doing nothing. Nothing in this repo references them any more (I grepped), so the risk is entirely in the deployment repos and in operators' own env files.

Given the PR description already flags a release-note item for the drop-timing change, the removed env vars deserve a line there too.

dmytro-landiak (commented):
## Review summary

Reviewed 45 changed files in [feat(limits): cluster-wide total throughput quota charged before persistence](https://github.com/thingsboard/tbmq/pull/356). Left 29 comment(s) inline.

The quota state machine in `ThroughputQuotaServiceImpl` holds up under scrutiny — I traced the credit floor, the dry backoff, the fail-open window and the stale-success guard against each other and could not find a state where a charge is granted twice or a deficit is never repaid. Same for the callback accounting in `MsgPersistenceManagerImpl`: `callbackCount` is computed over the full fan-out and every quota-refused subscription settles through `settleQuotaDropped`, so the wrapper always reaches zero. I also checked the ingress path for a `ByteBuf` leak now that the refusal branch no longer builds an `MqttPublishMsg` — `extractPublishMsg` is the only thing that retains and `channelRead` still `safeRelease`s the original, so that is clean.

Most of what follows is quality, testing and operability. Two things are worth a closer look before merge: the retained-message truncation comment doesn't hold for `retainHandling = 1`, and the non-idempotent lease return can inflate the cluster budget on a Redis timeout.

## Additional findings

These observations are about code outside this repo's diff — spotted while reading surrounding context.

- **Deployment repos** — `MQTT_RATE_LIMITS_THREADS_COUNT` / `_BATCH_SIZE` / `_PERIOD_MS` are gone from the broker yml (see the inline comment on that line). Nothing in this repo still references them, but `tbmq-pe-k8s`, `tbmq-pe-docker-compose` and `helm-charts` may still set them, and the three new keys (`block-size`, `lease-return-ms`, plus the reworded `config`) need propagating there. Worth a sweep with the config-propagation checklist before this ships.

---

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/persistence/MsgPersistenceManagerImpl.java
```diff
@@ -89,74 +91,107 @@ public void processPublish(PublishMsgWithId publishMsgWithId, PersistentMsgSubsc
...
-                processDeviceSubscriptionsWithRateLimits(deviceSubscriptions, publishMsgWithId, callbackWrapper);
-            } else {
-                processDeviceSubscriptions(deviceSubscriptions, publishMsgWithId, callbackWrapper);
-            }
+            // the narrower device-persisted-msgs limit truncates first; the total quota then pays only for what
+            // actually gets stored, so it never bills for messages that limit already dropped
+            List<Subscription> admittedDeviceSubscriptions = rateLimitService.isDevicePersistedMsgsLimitEnabled()
+                    ? applyDevicePersistedMsgsLimit(deviceSubscriptions, callbackWrapper)
+                    : deviceSubscriptions;
+            processDeviceSubscriptionsWithThroughputQuota(admittedDeviceSubscriptions, publishMsgWithId, callbackWrapper);
```
dmytro-landiak, line 100: A publish that fans out to devices, apps, shared topics and integrations now issues four separate charges, each of which can independently fall short and trigger its own synchronous Redis draw on the dispatcher thread — up to four round trips for one message. Charging once for the combined fan-out width and then apportioning the grant across the four groups would be one draw, and it would also make the truncation policy explicit rather than an accident of the call order (today devices always win over integrations because they are charged first).

Was the per-group split a deliberate prioritisation choice, or just how the code fell out?

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/persistence/MsgPersistenceManagerImpl.java
```diff
@@ -89,74 +91,107 @@ public void processPublish(PublishMsgWithId publishMsgWithId, PersistentMsgSubsc
...
-            log.trace("No available tokens left for device persisted messages bucket. Dropping {} messages", totalCount);
-            callbackWrapper.onBatchSuccess(totalCount);
-            return;
+    void processApplicationSubscriptionsWithThroughputQuota(List<Subscription> applicationSubscriptions,
+                                                            PublishMsgWithId publishMsgWithId,
+                                                            PublishMsgCallback callbackWrapper) {
+        int totalCount = applicationSubscriptions.size();
+        int granted = throughputQuotaService.tryConsumeOutgoingWaiting(totalCount);
+        for (int i = 0; i < granted; i++) {
+            sendApplicationMsg(applicationSubscriptions.get(i), publishMsgWithId, callbackWrapper);
```
dmytro-landiak, line 150: `processDeviceSubscriptionsWithThroughputQuota`, `processApplicationSubscriptionsWithThroughputQuota` and `processIntegrationSubscriptionsWithThroughputQuota` are the same six lines three times over, differing only in which `send*Msg` is invoked. A single `chargeAndSend(List<Subscription>, sender, publishMsgWithId, callbackWrapper)` taking a functional sender would collapse them — and if per-publish lambda allocation on this hot path is a concern, three `private final` method-reference fields initialised once in the constructor avoid it entirely. As it stands, a change to the charge/settle protocol has to be applied in three places and it is easy to miss one.

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/persistence/MsgPersistenceManagerImpl.java
```diff
@@ -89,74 +91,107 @@ public void processPublish(PublishMsgWithId publishMsgWithId, PersistentMsgSubsc
...
+        // a view over the caller's list, not a copy: safe because nothing downstream mutates either one
+        return availableTokens <= 0 ? List.of() : deviceSubscriptions.subList(0, availableTokens);
+    }
+
+    void processDeviceSubscriptionsWithThroughputQuota(List<Subscription> deviceSubscriptions,
+                                                       PublishMsgWithId publishMsgWithId,
+                                                       PublishMsgCallback callbackWrapper) {
+        int totalCount = deviceSubscriptions.size();
+        if (totalCount == 0) {
+            return; // the device limit took everything: nothing left to charge for
```
dmytro-landiak, line 135: Only the device path has this zero guard; the application, shared-topic and integration variants happily call `tryConsumeOutgoingWaiting(0)`. It works — the service grants a non-positive `n` in full — but the asymmetry invites the reader to wonder what is special about devices. If the guard belongs anywhere it is inside a shared charge helper, so all four sites behave the same way.

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/persistence/MsgPersistenceManagerImpl.java
```diff
@@ -89,74 +91,107 @@ public void processPublish(PublishMsgWithId publishMsgWithId, PersistentMsgSubsc
...
-                callbackWrapper.onSuccess();
+    void processSharedTopicsWithThroughputQuota(Set<String> sharedTopics,
+                                                PublishMsgWithId publishMsgWithId,
+                                                PublishMsgCallback callbackWrapper) {
+        PublishMsgProto publishMsgProto = publishMsgWithId.getPublishMsgProto();
+        int totalCount = sharedTopics.size();
+        int granted = throughputQuotaService.tryConsumeOutgoingWaiting(totalCount);
+        int sent = 0;
+        for (String sharedTopic : sharedTopics) {
+            if (sent == granted) {
```
dmytro-landiak: The manual `sent` counter plus mid-loop break is a slightly awkward way to take a prefix of a `Set`. An explicit `Iterator` with `for (int i = 0; i < granted && it.hasNext(); i++)` would read closer to the index loops in the three sibling methods, which would also make the family easier to unify. Minor, but it is the one method in the group that looks different for a reason that is not semantic.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
```diff
@@ -0,0 +1,334 @@
...
+            if (available <= 0) {
+                if (scheduleDraw) {
+                    scheduleDrawIfNeeded(n);
+                }
+                // the fail-open window above lapses while the next draw is still hanging on an unreachable
+                // Redis, so without this the quota would turn fail-closed for that draw's whole duration
+                return redisDegraded ? n : 0;
+            }
+            int granted = (int) Math.min(n, available);
+            if (localTokens.compareAndSet(current, current - granted)) {
```
dmytro-landiak, line 207: Every PUBLISH — ingress plus each fan-out group — CASes this one `AtomicLong` and calls `System.nanoTime()` at least once. At the msg/s figures this broker targets, that is a single cache line ping-ponging across every Netty event loop and dispatcher thread on the node, and CAS retries under contention make it worse than the uncontended cost suggests.

It is opt-in so the default path pays nothing, which takes the urgency off. Has it been measured with the quota on at target throughput? A striped pool (per-thread or per-core sub-leases refilled from the node lease) would keep the hot path thread-local, at the cost of coarser draw granularity.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
```diff
@@ -0,0 +1,334 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+
+@Slf4j
+@Service
+@RequiredArgsConstructor
+public class ThroughputQuotaServiceImpl implements ThroughputQuotaService {
+
+    static final long DRY_BACKOFF_NANOS = TimeUnit.MILLISECONDS.toNanos(50);
+    static final long FAIL_OPEN_NANOS = TimeUnit.SECONDS.toNanos(1);
```
dmytro-landiak, line 42: The fail-open window, dry backoff and clamp-warn interval are compile-time constants, which has two consequences: an operator hitting a flaky Redis cannot widen the 1s fail-open window without a rebuild, and the unit tests have to `Thread.sleep(1100)` and `Thread.sleep(60)` to cross them. Both would improve if these were `@Value` properties like `block-size` / `lease-return-ms` already are, or if the class took a nanoTime supplier the tests could advance. At minimum the fail-open window feels like something an operator would want to tune.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
```diff
@@ -0,0 +1,334 @@
...
+    private final StatsManager statsManager;
+
+    @Value("${mqtt.rate-limits.total.block-size:0}")
+    int configuredBlockSize;
+    @Value("${mqtt.rate-limits.total.lease-return-ms:1000}")
+    long leaseReturnMs;
+
+    // test seams: init() creates them only when still null; volatile so the caller threads that read
+    // them see the fully constructed executors and not a stale null
+    volatile ExecutorService drawExecutor;
```
dmytro-landiak, line 56: These two fields are package-private and mutable purely so tests can pre-assign them, with `init()` doing `if (drawExecutor == null)` to accommodate that — the comment calls them "test seams" outright. It works, but production initialisation is being shaped by the tests and nothing stops a future caller swapping the executor at runtime. Constructor-injecting an executor factory (or an `ExecutorService` provider) would give the tests the same control without leaving mutable lifecycle state on the bean.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaService.java
```diff
@@ -0,0 +1,81 @@
...
+     * RETAINED DELIBERATELY, despite having no production caller since the fan-out charges moved to
+     * {@link #tryConsumeOutgoingWaiting(int)}: it is the only non-blocking bulk charge in this API — the method
+     * to reach for if the retained-delivery path should ever stop blocking the subscription-persist I/O thread —
+     * and {@code ThroughputQuotaServiceImplTest} uses it to drain the local pool while asserting something else.
+     * Not dead code to be swept.
+     *
+     * @return the granted count in {@code [0..n]}; callers deliver the granted prefix and
+     * terminally settle the remainder
+     */
+    int tryConsumeOutgoing(int n);
```
dmytro-landiak, line 53: The javadoc openly says this has no production caller and argues against deleting it, which is a fairly strong signal it should not be on the interface. The two reasons given do not quite hold up: "the method to reach for if the retained path stops blocking" is speculative (reinstating a one-line delegate later is trivial), and "the test uses it to drain the local pool" means a public API method exists to serve a test, which is backwards.

Could the tests drain via repeated `tryConsumeOutgoing()` (or a package-private helper on the impl) and the interface shrink to the three methods production actually calls? A four-method interface where one is unused and two others are byte-identical (`tryConsume(1) == 1`) is more surface than this abstraction needs.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaService.java
```diff
@@ -0,0 +1,81 @@
...
+     *     <li>retained delivery on SUBSCRIBE, which runs on the {@code client-subscriptions} Kafka producer's
+     *     single I/O thread (the persist callback). A draw there also delays other clients' SUBACKs, which is
+     *     accepted because it is ONE bulk draw per retained-matching SUBSCRIBE rather than one per message,
+     *     and a Redis failure opens the fail-open window that makes this method return immediately.</li>
+     * </ul>
+     *
+     * @return the granted count in {@code [0..n]}; a remainder means the shared bucket is genuinely dry, so
+     * callers settle it terminally
+     */
+    int tryConsumeOutgoingWaiting(int n);
```
dmytro-landiak: "Waiting" does not convey the thing a caller most needs to know: this method can perform a synchronous Redis round trip on your thread. The javadoc has to spend a paragraph saying "never call this from the Netty ingress path" precisely because the name does not warn you. Something like `tryConsumeOutgoingBlocking` or `tryConsumeOutgoingWithSyncDraw` would make the hazard visible at the call site, where the reader is, rather than only in the javadoc.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaService.java
```diff
@@ -0,0 +1,81 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.thingsboard.mqtt.broker.service.limits;
+
+/**
+ * Cluster-wide total throughput quota — the single policy owner for the billed unit: MQTT PUBLISH
+ * packets processed, incoming and outgoing combined; messages routed to integration subscribers
+ * count as outgoing (spec: docs/superpowers/specs/2026-08-12-prepaid-quota-charging-design.md).
```
dmytro-landiak: This path does not exist in the repo, and `/docs/` is gitignored (widened in this very PR), so it never will. A future reader — or a PE engineer porting this — will hunt for a document they cannot reach. Either inline the two or three sentences of design rationale that actually matter, or point at something durable like the PR number.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputLimitProvider.java
```diff
@@ -15,12 +15,13 @@
...
 
-import org.thingsboard.mqtt.broker.actors.client.messages.mqtt.MqttPublishMsg;
-
-import java.util.function.Consumer;
-
-public interface RateLimitBatchProcessor {
-
-    void addMessage(MqttPublishMsg message, Consumer<MqttPublishMsg> onSuccess, Consumer<MqttPublishMsg> onRateLimits);
+/**
+ * Edition seam for the total throughput quota (spec §4.3). CE derives the sustained rate from the
```
dmytro-landiak: Same as the `ThroughputQuotaService` javadoc: "spec §4.3" refers to a document that is not in the repo and cannot be, since `/docs/` is gitignored. Either state the rule inline (CE derives from the configured bandwidths, PE overrides with the licensed rate — which the next sentence already does) or drop the citation.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/DefaultThroughputLimitProvider.java
```diff
@@ -0,0 +1,40 @@
...
+public class DefaultThroughputLimitProvider implements ThroughputLimitProvider {
+
+    private final TotalMsgsRateLimitsConfiguration totalMsgsRateLimitsConfiguration;
+
+    @Override
+    public long getSustainedRatePerSec() {
+        long minRatePerSec = Long.MAX_VALUE;
+        for (String limitSrc : totalMsgsRateLimitsConfiguration.getConfig().split(BrokerConstants.COMMA)) {
+            String[] parts = limitSrc.split(BrokerConstants.COLON);
+            long capacity = Long.parseLong(parts[0]);
```
dmytro-landiak: This re-implements the `capacity:seconds` split-and-parse that `AbstractMsgsRateLimitsConfiguration#getBucketConfiguration` already does, minus its `parts.length != 2` validation — so a malformed config produces a clean `IllegalArgumentException` down one path and an `ArrayIndexOutOfBoundsException` down this one (and a `:0` bandwidth divides by zero here).

Either derive the rate from `getBucketConfiguration().getBandwidths()` — each `Bandwidth` carries capacity and refill period — or lift the parsing into one shared helper on the config class that both callers use.

application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/handlers/MqttSubscribeHandler.java
```diff
@@ -213,24 +213,22 @@ private void processRetainedMessages(ClientSessionCtx ctx,
...
-
         int totalMsgCount = retainedMsgList.size();
-        int availableTokens = (int) rateLimitService.tryConsumeTotalMsgs(totalMsgCount);
-
-        if (availableTokens >= totalMsgCount) {
+        // wait for a right-sized draw: a retained set is one bulk charge, so the plain charge would cap the whole
+        // set at the node-local pool plus one block and silently truncate a subscriber that the cluster had budget
+        // for. See tryConsumeOutgoingWaiting for why blocking this thread (the subscription-persist callback) is
+        // acceptable here.
+        int granted = throughputQuotaService.tryConsumeOutgoingWaiting(totalMsgCount);
```
dmytro-landiak: This runs on the `client-subscriptions` Kafka producer's single I/O thread, so a Redis round trip here stalls every other client's SUBACK on the node, not just this one. The javadoc acknowledges it and argues it is one draw per retained-matching SUBSCRIBE — fair for a handful of retained topics, but a deployment with a busy retain store and many wildcard subscribers turns that thread into a serialisation point whose latency is now a function of Redis RTT.

Would it be feasible to hand the retained delivery off to a small dedicated executor before charging, so the producer callback thread never blocks?

application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/handlers/MqttSubscribeHandler.java
```diff
@@ -213,24 +213,22 @@ private void processRetainedMessages(ClientSessionCtx ctx,
     }
 
     List<RetainedMsg> applyRateLimits(List<RetainedMsg> retainedMsgList) {
```
dmytro-landiak: The name is now stale — this applies only the throughput quota, and every log line inside it already says "quota". `applyThroughputQuota` would match the body and the rest of the PR's vocabulary; the four test method names would follow.

application/src/main/java/org/thingsboard/mqtt/broker/server/MqttSessionHandler.java
```diff
@@ -242,26 +241,17 @@ private void processPublish(MqttMessage msg) {
...
-        MqttPublishMsg mqttPublishMsg = NettyMqttConverter.createMqttPublishMsg(sessionId, publishMsg);
-
-        if (!rateLimitService.isTotalMsgsLimitEnabled()) {
-            clientMqttActorManager.processMqttMsg(clientId, mqttPublishMsg);
+        if (!throughputQuotaService.tryConsumeIncoming()) {
+            tbMessageStatsReportClient.reportDroppedMsgs();
+            processMsgOnRateLimits(
+                    publishMsg.variableHeader().packetId(),
+                    publishMsg.fixedHeader().qosLevel().value(),
+                    TOTAL_RATE_LIMITS_DETECTED
```
dmytro-landiak, line 249: The new quota refusal block is a verbatim copy of the client-limit refusal block directly above it, down to the `publishMsg.variableHeader().packetId()` / `fixedHeader().qosLevel().value()` extraction. A small `refusePublish(publishMsg, reason)` helper would leave two readable guard clauses and one place to fix if the refusal protocol ever changes.

application/src/main/java/org/thingsboard/mqtt/broker/service/processing/downlink/basic/BasicDownLinkProcessorImpl.java
```diff
@@ -49,6 +51,10 @@ public void process(String clientId, PublishMsgProto msg) {
             dropMessage();
             return;
         }
+        if (!throughputQuotaService.tryConsumeOutgoing()) {
```
dmytro-landiak, line 54: This is the charge site for ordinary non-persistent (QoS 0 / clean-session) delivery — the most common delivery path in most deployments — and it is the only one of the six charge sites with no integration test; the other five each have a dedicated `TotalThroughputQuota*IntegrationTestCase`.

It looks entirely feasible to cover: clean-session subscribers, drain the shared bucket through `RateLimitCacheService` exactly as the fan-out test does, publish, then assert `delivered + droppedMsgs == subscribers`. The two new unit tests pin the branch but not that the drop is actually reported and accounted for end to end.

Worth noting this site also behaves differently from the persistent fan-out: it uses the lease-capped `tryConsumeOutgoing()`, so a burst can be dropped because the *node's* pool momentarily ran dry even while the cluster still has budget. That is defensible for best-effort QoS 0, but it is an asymmetry an integration test would document.

application/src/main/resources/thingsboard-mqtt-broker.yml
```diff
@@ -1008,20 +1008,41 @@ mqtt:
...
+      # per second sustained.
+      # IMPORTANT: the capacity is also the largest amount a SINGLE publish can charge, and when more than one bandwidth
+      # is configured the smallest available capacity applies. With the default, a publish routed to more than 1000
+      # subscribers is therefore always truncated, even on an idle cluster. If any topic in your deployment fans out to
+      # more subscribers than that, configure a single bandwidth whose capacity exceeds your widest fan-out —
+      # 50000:60 keeps the same ~833 messages per second sustained while allowing a 50000-subscriber fan-out.
       # Applied on broker start. Changing this value resets the token counters across the cluster when more than one
-      # limit:seconds pair is configured, granting one full interval of fresh budget.
+      # capacity:seconds bandwidth is configured, granting one full interval of fresh budget.
       config: "${MQTT_TOTAL_RATE_LIMITS_CONFIG:1000:1,50000:60}"
```
dmytro-landiak: The new comment above this line spends a whole IMPORTANT paragraph explaining that a multi-bandwidth config caps a single publish at the smallest capacity, that the shipped default therefore always truncates a fan-out above 1000 subscribers even on an idle cluster, and that operators should switch to a single bandwidth — and then ships exactly the config it warns about.

Since prepaid charging is what turned this from harmless into a per-publish ceiling, is there a reason not to make `50000:60` the default (same ~833 msg/s sustained, no sub-capacity per-publish cap) and let the paragraph shrink to a note about multi-bandwidth configs? Documenting around a footgun default is a worse outcome than removing it, especially as the quota is opt-in so nobody depends on the current default's exact shape yet.

application/src/main/resources/thingsboard-mqtt-broker.yml
```diff
@@ -1008,20 +1008,41 @@ mqtt:
...
+      # Number of quota tokens a broker node draws from Redis per replenishment round-trip. Defaults to 0, which derives
+      # max(1, sustainedRate / 10) from the config above. Larger blocks reduce Redis calls but let idle nodes hold more
+      # of the cluster budget between lease-return ticks. A publish whose fan-out exceeds the node's drawn tokens draws
+      # the shortfall on demand, so this value does not cap a single publish.
+      block-size: "${MQTT_TOTAL_RATE_LIMITS_BLOCK_SIZE:0}"
+      # Period of the background tick that returns a node's unused drawn tokens to the shared Redis bucket (in ms).
+      # Defaults to 1000. Set to 0 to disable the return tick.
+      lease-return-ms: "${MQTT_TOTAL_RATE_LIMITS_LEASE_RETURN_MS:1000}"
```
dmytro-landiak, line 1052: "Set to 0 to disable the return tick" states the mechanism but not the consequence: with the tick off, a node that draws a block and then goes quiet holds those tokens indefinitely, so the cluster's effective budget shrinks by the sum of every idle node's lease. That is a real footgun in a cluster with uneven traffic, and worth one more sentence here — the `block-size` comment already does this well by explaining that larger blocks "let idle nodes hold more of the cluster budget between lease-return ticks".

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/persistence/application/ApplicationPersistenceProcessorImpl.java
```diff
@@ -72,6 +72,7 @@
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
```
dmytro-landiak: This import is the entire production diff for this file, and nothing in the class uses a bare `Map` — every usage is `ConcurrentMap` / `ConcurrentHashMap`. It looks like a leftover from the delivery-time app-pack charging that was later reverted. Worth dropping so the file has no diff at all in this PR.

application/src/test/java/org/thingsboard/mqtt/broker/service/mqtt/persistence/application/ApplicationPersistenceProcessorImplTest.java
```diff
@@ -577,6 +577,12 @@ void processMainPack_retryAll_mustObtainAckStrategyOncePerPack_notOncePerRetry()
 
     // ===== Helpers =====
 
+    private PersistedPublishMsg publishMsg(int packetId, long offset) {
```
dmytro-landiak: This helper has no callers in the test class — same story as the unused `Map` import in the production file, i.e. residue from the reverted app-pack charging. Dead test scaffolding tends to get copied by the next person as if it were the house pattern; I would remove it.

application/src/test/java/org/thingsboard/mqtt/broker/service/limits/RateLimitRedisCacheServiceImplTest.java
```diff
@@ -305,4 +306,38 @@ public void givenLimitDisabledAndNoConfigurationInjected_whenInit_thenStartupSuc
...
+    public void testReturnTotalMsgs() {
+        RateLimitRedisCacheServiceImpl service = serviceWithTotalMsgsBucket();
+
+        service.returnTotalMsgs(25L);
+
+        verify(bucketProxy).addTokens(25L);
+    }
+
+    @Test
+    public void testReturnTotalMsgsWithNonPositiveCount() {
```
dmytro-landiak, line 319: The two new tests cover the `tokens <= 0` half of `returnTotalMsgs`'s guard but not the `totalMsgsBucketProxy == null` half — which is the quota-disabled shape, i.e. the default configuration. The service built by `setUp` already has a null proxy, so it is a two-line test: call `returnTotalMsgs(25)` on it and assert nothing blows up and no proxy is touched.

application/src/test/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImplTest.java
```diff
@@ -0,0 +1,396 @@
...
+        service.init();
+        executor.runAll(); // the warm-up draw lands: block 10 tokens local
+        clearInvocations(rateLimitCacheService);
+
+        int granted = service.tryConsumeOutgoingWaiting(50);
+
+        assertEquals("the pool plus one synchronous draw must cover the whole demand", 50, granted);
+        verify(rateLimitCacheService, times(1)).tryConsumeTotalMsgs(anyLong());
+        assertTrue("the waiting charge must not also queue an asynchronous draw", executor.tasks.isEmpty());
+        AtomicLong localTokens = (AtomicLong) ReflectionTestUtils.getField(service, "localTokens");
```
dmytro-landiak, line 293: Two assertions reach into private state by string name (`"localTokens"`, `"redisDegraded"`), so a field rename compiles fine and fails at runtime with an unhelpful message. Package-private accessors on the impl — the class already exposes `drawExecutor` and `draw` that way — would keep the same coverage and let the compiler protect the tests. Also `assertEquals(msg, true, ReflectionTestUtils.getField(...))` in the stale-success test would read better as `assertTrue`.

application/src/test/java/org/thingsboard/mqtt/broker/service/testing/integration/TotalThroughputQuotaAppPersistedIntegrationTestCase.java
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    private MqttConnectOptions getConnectOptions(boolean cleanSession, String username) {
+        MqttConnectOptions connectOptions = new MqttConnectOptions();
+        connectOptions.setCleanSession(cleanSession);
+        connectOptions.setUserName(username);
+        return connectOptions;
+    }
+
+    private double droppedMsgs() {
```
dmytro-landiak: Five of the six new classes carry an identical `droppedMsgs()`, four repeat `DRAIN_TOKENS = 10_000` plus the same drain-and-assert-it-drained sequence, three repeat the `awaitReceived` matcher helper, and all six repeat the same four-line `@TestPropertySource`. An `AbstractTotalThroughputQuotaIntegrationTest` holding the meter lookup, the drain helper and the shared properties would cut a lot of this and make the per-test ledger arithmetic — which is the genuinely interesting part — stand out.

application/src/test/java/org/thingsboard/mqtt/broker/service/testing/integration/TotalThroughputQuotaRetainedIntegrationTestCase.java
```diff
@@ -0,0 +1,186 @@
...
+        Thread.sleep(2000);
+        int truncated = receivedB.get();
+        assertTrue("a dry shared bucket must truncate the retained set, but got " + truncated,
+                truncated < RETAINED_COUNT);
+        assertEquals("retained truncation must NOT report droppedMsgs",
+                droppedBeforeTruncation, droppedMsgs(), 0.0);
+        disconnectAndClose(subB);
+
+        // phase 3 - the truncated messages were never consumed: after a refill the store still serves all five
+        Thread.sleep(6000);
```
dmytro-landiak: The six new classes add roughly 29 seconds of unconditional `Thread.sleep` on top of six separate Spring context boots (each class is its own `@ContextConfiguration` with `@DirtiesContext`). Some of these sleeps are unavoidable — a "nothing more arrives" assertion needs a fixed wait — but the positive waits ("let the backlog land", "let the dispatcher finish") could poll with Awaitility on the same counters the tests already track, which is both faster and less prone to going flaky on a loaded CI box. Worth checking whether a couple of these classes can share a context rather than each paying a full boot.

application/src/test/java/org/thingsboard/mqtt/broker/service/testing/integration/TotalThroughputQuotaRefusalIntegrationTestCase.java
```diff
@@ -0,0 +1,121 @@
...
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_CLASS)
+@ContextConfiguration(classes = TotalThroughputQuotaRefusalIntegrationTestCase.class, loader = SpringBootContextLoader.class)
+@TestPropertySource(properties = {
+        "mqtt.rate-limits.total.enabled=true",
+        "mqtt.rate-limits.total.config=5:600",
+        "mqtt.rate-limits.total.block-size=1",
+        "mqtt.rate-limits.total.lease-return-ms=600000"
+})
+@DaoSqlTest
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
```
dmytro-landiak: Test `b_` depends on test `a_` having drained the bucket, enforced by name-ordering and the `a_` / `b_` prefixes. That is fragile: running `b_` alone in an IDE fails, and a future rename silently reorders the suite. Since the other classes in this suite already drain the bucket explicitly through `RateLimitCacheService`, doing the same in a `@Before` here would make each test self-contained and let the prefixes and `@FixMethodOrder` go away.

.gitignore
```diff
@@ -36,6 +36,6 @@ rebuild-docker.sh
 */.run/**
 .run/**
 .run
-/docs/superpowers/
+/docs/
```
dmytro-landiak, line 39: Widening the ignore from `/docs/superpowers/` to the whole `/docs/` tree looks unrelated to the quota feature and probably should not ride along in this PR. It also has a side effect worth thinking about: nobody can commit real documentation under `docs/` again without noticing the ignore, and it is what makes the spec references in the new javadoc permanently unresolvable. If the intent is only to ignore local agent scratch space, a narrower pattern keeps `docs/` usable.

PR #358, opened 2026-08-20
fix(rate-limits): bound the throughput quota's fail-open window

dmytro-landiak:
## Pull Request description

The cluster-wide total throughput quota fails open when it cannot reach the shared Redis bucket. That is deliberate for a Redis blip, but the 1-second fail-open window was re-armed by every failed draw, so a longer outage never ended it: for as long as Redis stayed unreachable, every node granted every charge in full. Nothing was refused, nothing landed in `droppedMsgs`, and `throughputQuotaDegraded` only moved while traffic drove draws. Since Redis is required to start the broker but not to keep it running, stopping Redis after clients had connected bypassed the quota for an unbounded period.

This PR bounds the fail-open to one continuous outage:

- New `mqtt.rate-limits.total.degraded-grace-ms` (default `30000` — enough to ride out a Redis Cluster/Sentinel failover). The grace clock starts at the first failed draw of an outage and is not extended by later failures; any successful draw resets it, so an intermittent Redis never accumulates toward the deadline.
- Within the grace, behaviour is unchanged: charges are granted freely and the blip is absorbed.
- Past the grace the node lends nothing: the credit floor drops to 0, so it spends only tokens it actually drew and then refuses — reported in `droppedMsgs`; MQTT 5 publishers get `0x97 QUOTA_EXCEEDED`, MQTT 3.x publishers are disconnected.
- Refusing never stops probing: refusals still fall through to the draw path (limited to one probe per second) because a completed draw is the only thing that can notice Redis returning. The blocking bulk charge queues its probe on the draw executor instead of blocking an actor/consumer thread on the dead socket.
- The lease-return tick is skipped while degraded, so a hanging Redis call cannot strand the node's remaining tokens.
- The draw-failure WARN now says which way the quota is currently failing (open within the grace / refusing past it), and the quota javadocs were aligned with the bounded behaviour.
- The Redis-health state (degraded flag, outage start, last failure, fail-open window) is consolidated into one immutable `RedisHealth` snapshot swapped by CAS instead of four coordinated volatile fields, so the compound transition rules — a stale success must not clear a newer failure, the grace is stamped once per continuous outage — hold by construction rather than by write-ordering convention. The fail-open window is derived from the snapshot rather than stored, which removes the disarm-on-recovery bookkeeping.

Documentation notes: document `degraded-grace-ms`, and add a release note for the behaviour change — a Redis outage longer than the grace now refuses PUBLISH packets instead of passing unmetered traffic indefinitely.

Covered by unit tests: the grace ends fail-open; it measures one continuous outage (restarts after a successful draw, is not pushed out by repeated failures); the blocking charge refuses without a caller-thread draw but leaves a probe queued; enforcement resumes through the public charge methods once Redis answers; an MQTT 3.x publisher refused by the quota is disconnected without the packet reaching the actor; and the `RedisHealth` transitions are pinned deterministically (stale-success staleness rule, per-outage grace stamping, window disarm on recovery).

## General checklist

## Back-End feature checklist

### DIFF
--- a/application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
@@ -60,7 +60,10 @@ public class ThroughputQuotaServiceImpl implements ThroughputQuotaService {
     long degradedGraceMs;
 
     // Probe cadence while Redis is down. A seam like the executors below, so a test can shorten it instead of
-    // sleeping out the full interval.
+    // sleeping out the full interval - but unlike them deliberately NOT volatile: production never writes it after
+    // construction (init() leaves it alone), and a test assigns it on the same thread that then drives the charges,
+    // so program order already guarantees visibility. The executors are volatile because init() creates them at
+    // runtime and OTHER threads must see them fully constructed.
     long degradedProbeIntervalNanos = DEGRADED_PROBE_INTERVAL_NANOS;
 
     // test seams: init() creates them only when still null. Volatile so callers see fully constructed executors.
--- a/application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
@@ -168,7 +168,7 @@ public class ThroughputQuotaServiceImpl implements ThroughputQuotaService {
             scheduleDrawIfNeeded(blockSize);
             return granted;
         }
-        if (health.failOpenWindowActive(now)) {
+        if (health.failingOpen(now, degradedGraceNanos)) {
             return granted; // Redis degraded: tryConsume already failed open, nothing to draw
         }
         if (now - dryUntilNanos < 0) {
@@ -227,7 +227,7 @@ public class ThroughputQuotaServiceImpl implements ThroughputQuotaService {
         // window and on credit. Deliberately NOT an early return - the draw further down is the only thing that ever
         // notices Redis coming back, so short-circuiting here would make the refusal permanent.
         boolean graceExpired = health.graceExpired(now, degradedGraceNanos);
-        if (!graceExpired && health.failOpenWindowActive(now)) {
+        if (health.failingOpen(now, degradedGraceNanos)) {
             return n; // Redis is degraded but within the grace: fail open, never queue, never stall
         }
         // Grant on bounded credit while a draw may still help; once a draw confirmed the bucket dry, refuse locally
@@ -277,7 +277,7 @@ public class ThroughputQuotaServiceImpl implements ThroughputQuotaService {
         RedisHealth health = redisHealth.get();
         // Past the grace the pool lends nothing, so whatever the charge just spent was real budget and a drained
         // pool genuinely owes a refill: the fail-open window must not suppress the top-up there.
-        if (!health.graceExpired(now, degradedGraceNanos) && health.failOpenWindowActive(now)) {
+        if (health.failingOpen(now, degradedGraceNanos)) {
             return; // failing open spends no real budget, so nothing is owed
         }
         if (localTokens.get() <= 0) {
@@ -447,10 +447,20 @@ public class ThroughputQuotaServiceImpl implements ThroughputQuotaService {
             return healthy(now);
         }
 
+        /**
+         * Whether the quota is currently failing open: inside the short reprieve each failed draw arms, AND still
+         * inside the grace. The precedence lives here so no call site can consult the window alone - during an
+         * outage the probe cycle re-arms the window every few seconds, so window-without-grace is structurally the
+         * unbounded fail-open this deadline exists to end. That is why {@link #failOpenWindowActive} is private.
+         */
+        boolean failingOpen(long now, long graceNanos) {
+            return !graceExpired(now, graceNanos) && failOpenWindowActive(now);
+        }
+
         /** The short reprieve after each failed draw, derived instead of stored: while degraded it is always
          * lastFailure + FAIL_OPEN_NANOS, and recovery closes it because the flag drops - a stored copy existed
          * only to be disarmed in lockstep. */
-        boolean failOpenWindowActive(long now) {
+        private boolean failOpenWindowActive(long now) {
             // subtract rather than compare against a sum: these are nanoTime readings, so an added duration can
             // overflow and invert the test
             return degraded && now - lastFailureNanos < FAIL_OPEN_NANOS;
--- a/application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
@@ -291,11 +291,12 @@ public class ThroughputQuotaServiceImpl implements ThroughputQuotaService {
             // only at the call sites keeps a disabled service safe whatever reaches it.
             return;
         }
-        if (nanoTime.getAsLong() - dryUntilNanos < 0) {
+        long now = nanoTime.getAsLong();
+        if (now - dryUntilNanos < 0) {
             return; // bucket known dry: no Redis traffic during the backoff window
         }
         RedisHealth health = redisHealth.get();
-        if (health.degraded() && nanoTime.getAsLong() - health.lastFailureNanos() < degradedProbeIntervalNanos) {
+        if (health.degraded() && now - health.lastFailureNanos() < degradedProbeIntervalNanos) {
             return; // Redis is unreachable: one probe per interval, not one per refused packet
         }
         if (!drawInFlight.compareAndSet(false, true)) {
--- a/application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
@@ -33,6 +33,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.LongSupplier;
 
 @Slf4j
 @Service
@@ -66,6 +67,10 @@ public class ThroughputQuotaServiceImpl implements ThroughputQuotaService {
     // runtime and OTHER threads must see them fully constructed.
     long degradedProbeIntervalNanos = DEGRADED_PROBE_INTERVAL_NANOS;
 
+    // Clock seam with the same contract as the interval above: production never writes it, a test assigns it before
+    // driving traffic. It is what lets the grace and window arithmetic be tested without wall-clock sleeps.
+    LongSupplier nanoTime = System::nanoTime;
+
     // test seams: init() creates them only when still null. Volatile so callers see fully constructed executors.
     volatile ExecutorService drawExecutor;
     volatile ScheduledExecutorService leaseReturnScheduler;
@@ -92,7 +97,7 @@ public class ThroughputQuotaServiceImpl implements ThroughputQuotaService {
         if (!enabled) {
             return;
         }
-        long now = System.nanoTime();
+        long now = nanoTime.getAsLong();
         dryUntilNanos = now;
         redisHealth.set(RedisHealth.healthy(now));
 
@@ -154,7 +159,7 @@ public class ThroughputQuotaServiceImpl implements ThroughputQuotaService {
             repayCreditIfOwed();
             return granted;
         }
-        long now = System.nanoTime();
+        long now = nanoTime.getAsLong();
         RedisHealth health = redisHealth.get();
         if (health.graceExpired(now, degradedGraceNanos)) {
             // tryConsume has already refused on the deadline. Probe for recovery on the draw executor rather than
@@ -181,12 +186,12 @@ public class ThroughputQuotaServiceImpl implements ThroughputQuotaService {
         // over-draw lands in the pool, an under-draw leaves the rest for the next draw.
         long deficit = Math.max(0L, -localTokens.get());
         long drawSize = Math.max(blockSize, shortfall + deficit);
-        long startedAt = System.nanoTime();
+        long startedAt = nanoTime.getAsLong();
         try {
             long drawn = rateLimitCacheService.tryConsumeTotalMsgs(drawSize);
             markRedisAnswered(startedAt);
             if (drawn <= 0) {
-                dryUntilNanos = System.nanoTime() + DRY_BACKOFF_NANOS;
+                dryUntilNanos = nanoTime.getAsLong() + DRY_BACKOFF_NANOS;
                 warnQuotaClamped();
                 return 0;
             }
@@ -198,7 +203,7 @@ public class ThroughputQuotaServiceImpl implements ThroughputQuotaService {
             return granted;
         } catch (Exception e) {
             // fail open, consistent with tryConsume - including its deadline: bridge a blip, refuse an outage
-            return markRedisFailed(e).graceExpired(System.nanoTime(), degradedGraceNanos) ? 0 : shortfall;
+            return markRedisFailed(e).graceExpired(nanoTime.getAsLong(), degradedGraceNanos) ? 0 : shortfall;
         }
     }
 
@@ -215,7 +220,7 @@ public class ThroughputQuotaServiceImpl implements ThroughputQuotaService {
         if (!enabled || n <= 0) {
             return Math.max(0, n);
         }
-        long now = System.nanoTime();
+        long now = nanoTime.getAsLong();
         // one snapshot for the whole charge, so the grace, the window and the degraded flag cannot disagree mid-call
         RedisHealth health = redisHealth.get();
         // Redis has been unreachable long enough that this is an outage, not a blip: stop granting on the fail-open
@@ -268,7 +273,7 @@ public class ThroughputQuotaServiceImpl implements ThroughputQuotaService {
     // The blocking charge suppresses tryConsume's scheduling, so when it returns without drawing it must put the
     // top-up back: a pool at or below zero owes the shared bucket for the credit it granted.
     private void repayCreditIfOwed() {
-        long now = System.nanoTime();
+        long now = nanoTime.getAsLong();
         RedisHealth health = redisHealth.get();
         // Past the grace the pool lends nothing, so whatever the charge just spent was real budget and a drained
         // pool genuinely owes a refill: the fail-open window must not suppress the top-up there.
@@ -286,11 +291,11 @@ public class ThroughputQuotaServiceImpl implements ThroughputQuotaService {
             // only at the call sites keeps a disabled service safe whatever reaches it.
             return;
         }
-        if (System.nanoTime() - dryUntilNanos < 0) {
+        if (nanoTime.getAsLong() - dryUntilNanos < 0) {
             return; // bucket known dry: no Redis traffic during the backoff window
         }
         RedisHealth health = redisHealth.get();
-        if (health.degraded() && System.nanoTime() - health.lastFailureNanos() < degradedProbeIntervalNanos) {
+        if (health.degraded() && nanoTime.getAsLong() - health.lastFailureNanos() < degradedProbeIntervalNanos) {
             return; // Redis is unreachable: one probe per interval, not one per refused packet
         }
         if (!drawInFlight.compareAndSet(false, true)) {
@@ -313,14 +318,14 @@ public class ThroughputQuotaServiceImpl implements ThroughputQuotaService {
     }
 
     void draw(long drawSize) {
-        long startedAt = System.nanoTime();
+        long startedAt = nanoTime.getAsLong();
         try {
             long granted = rateLimitCacheService.tryConsumeTotalMsgs(drawSize);
             markRedisAnswered(startedAt);
             if (granted > 0) {
                 localTokens.addAndGet(granted); // repays any credit deficit first
             } else {
-                dryUntilNanos = System.nanoTime() + DRY_BACKOFF_NANOS;
+                dryUntilNanos = nanoTime.getAsLong() + DRY_BACKOFF_NANOS;
                 warnQuotaClamped();
             }
         } catch (Exception e) {
@@ -335,12 +340,12 @@ public class ThroughputQuotaServiceImpl implements ThroughputQuotaService {
     // it armed would grant every charge in full for the rest of the second, unmetered traffic at the exact moment
     // enforcement should resume.
     private void markRedisAnswered(long drawStartedNanos) {
-        long now = System.nanoTime();
+        long now = nanoTime.getAsLong();
         redisHealth.updateAndGet(health -> health.answered(drawStartedNanos, now));
     }
 
     private RedisHealth markRedisFailed(Exception e) {
-        long now = System.nanoTime();
+        long now = nanoTime.getAsLong();
         RedisHealth health = redisHealth.updateAndGet(h -> h.failed(now));
         stats.incrementRedisDegraded();
         // Say which way the quota is currently failing. Claiming "failing open" while the node is in fact refusing
@@ -357,7 +362,7 @@ public class ThroughputQuotaServiceImpl implements ThroughputQuotaService {
     }
 
     private void warnQuotaClamped() {
-        long now = System.nanoTime();
+        long now = nanoTime.getAsLong();
         // an exhausted bucket can bring every delivery thread through here at once, so CAS the timestamp: exactly
         // one of them per interval wins the right to log
         long last = lastClampWarnNanos.get();
--- a/application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
+++ b/application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
@@ -41,6 +41,10 @@ public class ThroughputQuotaServiceImpl implements ThroughputQuotaService {
     static final long DRY_BACKOFF_NANOS = TimeUnit.MILLISECONDS.toNanos(50);
     static final long FAIL_OPEN_NANOS = TimeUnit.SECONDS.toNanos(1);
     static final long CLAMP_WARN_INTERVAL_NANOS = TimeUnit.SECONDS.toNanos(60);
+    // Past the grace every refused charge would otherwise queue a fresh draw, and drawInFlight bounds concurrency,
+    // not frequency - against a Redis that refuses connections instantly that is one round trip and one WARN per
+    // packet. Probe on an interval instead.
+    static final long DEGRADED_PROBE_INTERVAL_NANOS = TimeUnit.SECONDS.toNanos(1);
 
     private final TotalMsgsRateLimitsConfiguration totalMsgsRateLimitsConfiguration;
     private final ThroughputLimitProvider throughputLimitProvider;
@@ -51,6 +55,8 @@ public class ThroughputQuotaServiceImpl implements ThroughputQuotaService {
     int configuredBlockSize;
     @Value("${mqtt.rate-limits.total.lease-return-ms:1000}")
     long leaseReturnMs;
+    @Value("${mqtt.rate-limits.total.degraded-grace-ms:30000}")
+    long degradedGraceMs;
 
     // test seams: init() creates them only when still null. Volatile so callers see fully constructed executors.
     volatile ExecutorService drawExecutor;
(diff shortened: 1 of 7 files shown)


application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
```diff
@@ -317,6 +365,12 @@ private void warnQuotaClamped() {
     }
 
     void returnUnusedTokens() {
+        if (redisHealth.get().degraded()) {
```
dmytro-landiak, line 377: The degraded skip here has no test — the existing `returnUnusedTokens` tests cover only the healthy positive-balance and credit-deficit paths. A test that marks the node degraded with a positive pool and verifies `returnTotalMsgs` is never called would pin it cheaply, and the same fixture could pin the other half of what this skip protects: a grace-expired node still granting from tokens it actually owns. Every grace-expired test today starts from an empty pool, so the `creditFloor = 0` "spend only what was drawn" behavior is never exercised with a non-empty pool.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
```diff
@@ -230,7 +265,11 @@ private int tryConsume(int n, boolean scheduleDraw) {
     // The blocking charge suppresses tryConsume's scheduling, so when it returns without drawing it must put the
     // top-up back: a pool at or below zero owes the shared bucket for the credit it granted.
     private void repayCreditIfOwed() {
-        if (System.nanoTime() - failOpenUntilNanos < 0) {
+        long now = System.nanoTime();
+        RedisHealth health = redisHealth.get();
+        // Past the grace the pool lends nothing, so whatever the charge just spent was real budget and a drained
+        // pool genuinely owes a refill: the fail-open window must not suppress the top-up there.
+        if (!health.graceExpired(now, degradedGraceNanos) && health.failOpenWindowActive(now)) {
```
dmytro-landiak: The grace-expired side of this guard is untested. It exists so a blocking charge that fully grants from owned tokens past the grace still schedules the top-up when the pool drains — but no test drives `tryConsumeOutgoingBlocking` to a full grant while grace-expired (pool pre-loaded, then drained to zero) and asserts the draw gets scheduled. The probe-interval seam should make that deterministic to write.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
```diff
@@ -81,10 +91,14 @@ public void init() {
         }
         long now = System.nanoTime();
         dryUntilNanos = now;
-        failOpenUntilNanos = now;
-        lastDrawFailureNanos = now; // no failure yet, so every draw from here on is newer than this
+        redisHealth.set(RedisHealth.healthy(now));
 
         lastClampWarnNanos.set(now - CLAMP_WARN_INTERVAL_NANOS - 1); // so the first clamp warns immediately
+        if (degradedGraceMs < 0) {
```
dmytro-landiak, line 105: Neither boundary of this clamp is tested: nothing sets a negative `degradedGraceMs` and asserts it behaves as 0 (plus this WARN), and the yml's promise that `0` refuses the moment a draw fails has no test either. The zero case is the more interesting one — it only holds because `graceExpired` uses `>=`, which is exactly the kind of detail a later refactor to `>` would silently break.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
```diff
@@ -284,25 +327,30 @@ void draw(long drawSize) {
...
-        redisDegraded = true;
-        failOpenUntilNanos = now + FAIL_OPEN_NANOS;
+        RedisHealth health = redisHealth.updateAndGet(h -> h.failed(now));
         stats.incrementRedisDegraded();
-        log.warn("Failed to draw total throughput quota tokens from Redis, failing open for {} ms",
-                TimeUnit.NANOSECONDS.toMillis(FAIL_OPEN_NANOS), e);
+        // Say which way the quota is currently failing. Claiming "failing open" while the node is in fact refusing
+        // every PUBLISH would point an operator at exactly the wrong cause during a total publish outage.
+        if (health.graceExpired(now, degradedGraceNanos)) {
+            log.warn("Failed to draw total throughput quota tokens from Redis; the {} ms degraded grace has elapsed, " +
```
dmytro-landiak, line 355: The comment justifies the two wordings as operator-facing, but neither branch selection is asserted. The class already has the `ListAppender` pattern (`givenRepeatedDryDraws_whenClamping_thenWarnsOncePerInterval`), so pinning that a past-grace failure logs the "refused" wording rather than "failing open" would be a few lines in the existing style.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
```diff
@@ -247,6 +286,10 @@ private void scheduleDrawIfNeeded(long shortfall) {
         if (System.nanoTime() - dryUntilNanos < 0) {
             return; // bucket known dry: no Redis traffic during the backoff window
         }
+        RedisHealth health = redisHealth.get();
+        if (health.degraded() && System.nanoTime() - health.lastFailureNanos() < degradedProbeIntervalNanos) {
```
dmytro-landiak: This adds a second `System.nanoTime()` call two lines below the dry-backoff check's own; every other method in the file hoists a single `long now` for all its clock comparisons. Hoisting would match the house pattern and shave a clock read from what becomes a per-refused-packet path during an outage.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
```diff
@@ -345,4 +399,59 @@ int deriveDefaultBlockSize() {
...
+            if (lastFailureNanos - drawStartedNanos >= 0) {
+                return this; // stale success
+            }
+            return healthy(now);
+        }
+
+        /** The short reprieve after each failed draw, derived instead of stored: while degraded it is always
+         * lastFailure + FAIL_OPEN_NANOS, and recovery closes it because the flag drops - a stored copy existed
+         * only to be disarmed in lockstep. */
+        boolean failOpenWindowActive(long now) {
```
dmytro-landiak: Every call site now has to remember that grace-expiry trumps the window — the `!graceExpired(...) && failOpenWindowActive(...)` pairing is re-encoded three times (`tryConsume`, `repayCreditIfOwed`, and as sequential checks in `tryConsumeOutgoingBlocking`), and a future call site that consults the window alone silently reintroduces the unbounded fail-open this PR fixes. Would a combined predicate on the record — say `failingOpen(now, graceNanos)` — be worth it so the precedence lives in one place? `tryConsume` still needs the standalone `graceExpired` for the credit floor and the refusal, so it can't fully collapse there, but the other two sites would reduce to one call each.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
```diff
@@ -51,6 +56,12 @@ public class ThroughputQuotaServiceImpl implements ThroughputQuotaService {
     int configuredBlockSize;
     @Value("${mqtt.rate-limits.total.lease-return-ms:1000}")
     long leaseReturnMs;
+    @Value("${mqtt.rate-limits.total.degraded-grace-ms:30000}")
+    long degradedGraceMs;
+
+    // Probe cadence while Redis is down. A seam like the executors below, so a test can shorten it instead of
+    // sleeping out the full interval.
+    long degradedProbeIntervalNanos = DEGRADED_PROBE_INTERVAL_NANOS;
```
dmytro-landiak, line 68: Unlike the executor seams below, this one isn't volatile. That looks fine — tests write it before any traffic on the same thread, and production never writes it at all — but since "a seam like the executors below" invites reading it as the same visibility contract, a word on why volatility isn't needed here would save the next reader the analysis.

application/src/test/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImplTest.java
```diff
@@ -271,6 +275,118 @@ public void givenDryBucket_whenBackoffActive_thenRefusesLocallyAndRecoversAfterB
         assertTrue("after backoff, credit resumes and a new draw is scheduled", service.tryConsumeIncoming());
     }
 
+    // Fail-open is a bridge over a Redis blip, not a licence to serve unmetered traffic for as long as Redis stays
+    // down: the 1 s window is re-armed by every failed draw, so without a deadline on the CONTINUOUS outage the
+    // quota would never resume enforcing.
+    @Test
+    public void givenRedisDownPastGrace_whenConsuming_thenStopsGrantingFreely() throws InterruptedException {
```
dmytro-landiak: The integration family (`TotalThroughputQuota*IntegrationTestCase`) covers refusals on a spent budget, but the new outage semantics — grace expiry against an unreachable Redis, then recovery — exist only at this mock level. A suite with a short `degraded-grace-ms` (and short Redis timeouts) that stops the Valkey container, publishes past the grace, asserts the refusal through the real `droppedMsgs` counter, then restarts Valkey and asserts enforcement resumes would exercise the headline behavior end to end. If Jedis timeout tuning or container lifecycle makes that impractical, a note somewhere that the integration route was considered (and why unit tests were the stopping point) would help the next person who asks.

application/src/test/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImplTest.java
```diff
@@ -271,6 +275,118 @@ public void givenDryBucket_whenBackoffActive_thenRefusesLocallyAndRecoversAfterB
...
+        service.drawExecutor = MoreExecutors.newDirectExecutorService();
+        service.degradedGraceMs = 100;
+        when(rateLimitCacheService.tryConsumeTotalMsgs(anyLong())).thenThrow(new RuntimeException("redis down"));
+
+        service.init(); // warm-up draw fails -> degraded, the grace starts
+
+        assertTrue("within the grace the quota must still fail open", service.tryConsumeIncoming());
+
+        Thread.sleep(150); // past the grace, while the fail-open window is still being re-armed
```
dmytro-landiak: Five of the new tests sleep against real `System.nanoTime` with tight margins: the "within the grace" assertTrue above flips if anything stalls the thread for more than 100 ms between `init()` and the assert (a GC pause on loaded CI), and `givenIntermittentRedis` has ~100 ms of slack after its final sleep inside a 200 ms grace. Since the PR already made the probe interval an injectable seam for exactly this reason, a nanoTime supplier seam (`LongSupplier` defaulting to `System::nanoTime`) would make the grace arithmetic deterministic and drop ~3.5 s of sleeps per run — or, cheaper, widen the graces so every assert has at least 500 ms of slack.

dmytro-landiak (commented):
## Review summary

Reviewed 6 changed files in [fix(rate-limits): bound the throughput quota's fail-open window](https://github.com/thingsboard/tbmq/pull/358). Left 9 comment(s) inline.

No correctness issues found: the `RedisHealth` CAS transitions hold the rules they document (stale-success staleness, one grace stamp per continuous outage, window disarmed by recovery), the grace-over-window precedence is applied consistently at every read site, the deliberately-not-early-return in `tryConsume` plus the balance-crossing scheduling keeps recovery probing alive in every state I could construct, and the new handler test matches the real refusal path. The inline comments are test-coverage gaps and small design/readability questions.

---

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
dmytro-landiak: Added in 1ab7c99c6: `givenDegradedNodeWithOwnedTokens_whenReturnTick_thenTokensAreKeptLocal` pins the skip — degraded node, pool pre-loaded to 25, the tick calls `returnTotalMsgs` never and the tokens stay local. The other half is `givenGraceExpiredWithOwnedTokens_whenConsuming_thenSpendsOnlyWhatWasDrawn`: a grace-expired node with 5 owned tokens grants exactly those 5 and not one packet of credit, then refuses. Both use a zero grace ("expired from the first failure"), so neither sleeps.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
dmytro-landiak: Added in 1ab7c99c6: `givenGraceExpired_whenBlockingChargeGrantsFromOwnedTokens_thenTheDrainedPoolSchedulesTheTopUp` — pool pre-loaded to 30, `tryConsumeOutgoingBlocking(30)` fully grants past the grace while the fail-open window is freshly armed, and the drained pool queues exactly one draw on the executor with no caller-thread Redis call. Deterministic as suggested: `degradedProbeIntervalNanos = 0` so the probe gate cannot swallow the scheduling under test, and a zero grace removes the wall-clock wait.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
dmytro-landiak: Added in 1ab7c99c6: `givenNegativeGrace_whenInit_thenWarnsAndBehavesAsZero` asserts the startup WARN and that the first failed draw already refuses; `givenZeroGrace_whenTheFirstDrawFails_thenRefusesImmediately` covers the yml promise directly. The `>=` itself is pinned at the record level by `givenTheGraceBoundary_whenChecked_thenElapsedEqualToTheGraceIsExpired`: expired at exactly `elapsed == grace`, still inside one nano short — a refactor to `>` fails all three.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
dmytro-landiak: Fixed in 9c3601fe8 — the field comment now states the contract instead of inviting the analysis: production never writes it after construction (init() leaves it alone) and a test assigns it on the same thread that then drives the charges, so program order is visibility enough; the executors need volatile because init() creates them at runtime for other threads to see fully constructed.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
dmytro-landiak: Added in 9e7d699e6: `givenDrawFailures_whenWarning_thenTheWordingNamesTheCurrentFailureDirection` — the warm-up failure within the grace logs exactly one "failing open" WARN and zero "refused" ones; a second failing draw past the deadline (advanced via the new nanoTime seam, no sleeps) logs exactly one "refused until Redis answers" and does not add another "failing open".

application/src/test/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImplTest.java
dmytro-landiak: Took the seam route in 9e7d699e6: `LongSupplier nanoTime` (defaulting to `System::nanoTime`, same non-volatile contract as the probe-interval seam) now feeds every clock read, and the five grace tests advance an `AtomicLong` instead of sleeping — the class run dropped from ~5.6s to ~2.1s and the 100 ms margins are gone entirely. The pre-existing dry-backoff/window-lapse sleepers were left as they were.

application/src/test/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImplTest.java
dmytro-landiak: Added in 5d8ec2535: `TotalThroughputQuotaRedisOutageIntegrationTestCase` runs the full arc against a real Jedis failure — it PAUSES the Valkey container (blackholed socket, the harsh detection path; stop/start would remap the port under the built context), drives publishes until a draw actually fails, then asserts refusal through the real droppedMsgs counter plus a 0x97 PUBACK, unpauses, and asserts a publish is granted and stored again via a real probe draw. The budget is 600 tokens with ~1/s sustained, so no refusal in the suite can be a spent-budget one, and the grace is 0 — the grace DURATION arithmetic is pinned deterministically at the unit level against the new nanoTime seam, so the IT pins the wiring a mock cannot (Jedis timeout, client-visible refusal, recovery) without re-introducing wall-clock margins. Green locally in ~27s.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
dmytro-landiak: Fixed in b03c8a812 — a single `long now` now feeds both the dry-backoff check and the probe gate, matching the house pattern and dropping the second clock read from the per-refused-packet path.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
dmytro-landiak: Done in 9d2d1b3bf, one step further than proposed: `failingOpen(now, graceNanos)` carries the precedence, and `failOpenWindowActive` is now PRIVATE on the record — so a future call site cannot consult the window alone even by accident, rather than merely being offered a safer alternative next to the footgun. repayCreditIfOwed and the blocking path reduce to one call each; tryConsume also uses it (the internal graceExpired recompute is two comparisons on plain record fields) while keeping its standalone graceExpired for the credit floor and the refusal. The precedence itself is pinned by a new record test: an armed window with the grace over must not grant.

application/src/main/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImpl.java
```diff
@@ -230,7 +273,11 @@ private int tryConsume(int n, boolean scheduleDraw) {
     // The blocking charge suppresses tryConsume's scheduling, so when it returns without drawing it must put the
     // top-up back: a pool at or below zero owes the shared bucket for the credit it granted.
     private void repayCreditIfOwed() {
-        if (System.nanoTime() - failOpenUntilNanos < 0) {
+        long now = nanoTime.getAsLong();
+        RedisHealth health = redisHealth.get();
+        // Past the grace the pool lends nothing, so whatever the charge just spent was real budget and a drained
+        // pool genuinely owes a refill: the fail-open window must not suppress the top-up there.
+        if (health.failingOpen(now, degradedGraceNanos)) {
```
dmytro-landiak, line 280: This round added a test for this guard's false branch (grace expired → the drained pool schedules the top-up), but the true branch still has none: a blocking charge granted freely inside the armed window must NOT schedule a top-up, and no `tryConsumeOutgoingBlocking` test runs in that state — they cover disabled, healthy, and grace-expired. With the clock seam it's deterministic now (arm the window without advancing past the grace, charge, assert nothing was queued), and it would be the exact mirror of `givenGraceExpired_whenBlockingChargeGrantsFromOwnedTokens_thenTheDrainedPoolSchedulesTheTopUp`.

application/src/test/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImplTest.java
```diff
@@ -271,6 +275,265 @@ public void givenDryBucket_whenBackoffActive_thenRefusesLocallyAndRecoversAfterB
...
+        ListAppender<ILoggingEvent> appender = new ListAppender<>();
+        appender.start();
+        logger.addAppender(appender);
+        try {
+            service.init(); // the clamp warns, then the warm-up draw fails
+
+            assertEquals("a negative grace must be called out at startup", 1L,
+                    appender.list.stream()
+                            .filter(event -> event.getLevel() == Level.WARN)
+                            .filter(event -> event.getFormattedMessage().contains("degraded-grace-ms is negative"))
```
dmytro-landiak, line 475: This assert re-inlines the exact WARN-count stream that `warnsContaining` — added a few tests below in this same round — encapsulates; it could be `assertEquals(1L, warnsContaining(appender, "degraded-grace-ms is negative"))`. Relatedly, the Logger/ListAppender attach-try-detach boilerplate now appears three times in the file (this test, the WARN-wording test, and the pre-existing clamp test) — a small helper that runs a body against a captured appender would collapse all three.

application/src/test/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImplTest.java
```diff
@@ -271,6 +275,265 @@ public void givenDryBucket_whenBackoffActive_thenRefusesLocallyAndRecoversAfterB
...
+        assertFalse("the probing charge itself is still refused", service.tryConsumeIncoming());
+
+        // Recovered, so grants come from the drawn block plus the node's ordinary burst credit. Asking for far more
+        // than that is what separates recovery from a fail-open window: a node still failing open returns all 5000.
+        assertEquals("a node that could not probe Redis would never recover",
+                10 + service.deriveBurstAllowance(), service.tryConsumeOutgoing(5000));
+    }
+
+    private AtomicLong pool() {
+        return (AtomicLong) ReflectionTestUtils.getField(service, "localTokens");
```
dmytro-landiak, line 402: String-based reflection on `"localTokens"` breaks silently on a field rename, and the class under test already sets the precedent for the safer option: it exposes a package-private `redisHealth()` accessor for exactly these same-package tests. A package-private `localTokens()` accessor would make this compile-checked. And if the helper stays as is, `givenShortPool_whenBlockingCharge_thenOneDrawAndNoSecondQueued` still carries its own inline copy of the same reflection — worth at least pointing it at this helper.

application/src/test/java/org/thingsboard/mqtt/broker/service/limits/ThroughputQuotaServiceImplTest.java
```diff
@@ -271,6 +275,265 @@ public void givenDryBucket_whenBackoffActive_thenRefusesLocallyAndRecoversAfterB
         assertTrue("after backoff, credit resumes and a new draw is scheduled", service.tryConsumeIncoming());
     }
 
+    // Fail-open is a bridge over a Redis blip, not a licence to serve unmetered traffic for as long as Redis stays
+    // down: the 1 s window is re-armed by every failed draw, so without a deadline on the CONTINUOUS outage the
+    // quota would never resume enforcing.
+    @Test
+    public void givenRedisDownPastGrace_whenConsuming_thenStopsGrantingFreely() {
+        AtomicLong clock = new AtomicLong();
+        service.nanoTime = clock::get;
```
dmytro-landiak, line 284: The two-line fake-clock install is now copy-pasted into six tests. A tiny helper — `private AtomicLong installFakeClock()` that assigns the seam and returns the clock — would keep the next grace test from re-typing it.

dmytro-landiak (commented):
## Re-review summary

Re-reviewed [fix(rate-limits): bound the throughput quota's fail-open window](https://github.com/thingsboard/tbmq/pull/358) — verified 9 finding(s) from [previous review](https://github.com/thingsboard/tbmq/pull/358#pullrequestreview-4982450734).

| Status | Count |
|--------|-------|
| ✅ Resolved | 9 |
| 💬 Acknowledged | 0 |
| ❌ Unresolved | 0 |

All previous findings have been addressed — each fix was verified in the code, not just the reply. Also left 6 new comment(s) inline on the fix commits: no correctness issues, all test-coverage and cleanup suggestions.

### Finding details

- ✅ **ThroughputQuotaServiceImpl.java:377** — `returnUnusedTokens` degraded skip and grace-expired owned-token spending untested — *Fixed in code: `givenDegradedNodeWithOwnedTokens_whenReturnTick_thenTokensAreKeptLocal` pins the skip, `givenGraceExpiredWithOwnedTokens_whenConsuming_thenSpendsOnlyWhatWasDrawn` pins the creditFloor=0 grant from a non-empty pool.*
- ✅ **ThroughputQuotaServiceImpl.java:275** — `repayCreditIfOwed`'s grace-expired side untested — *Fixed in code: `givenGraceExpired_whenBlockingChargeGrantsFromOwnedTokens_thenTheDrainedPoolSchedulesTheTopUp` (deterministic: zero grace + zero probe interval).*
- ✅ **ThroughputQuotaServiceImpl.java:105** — neither boundary of the grace clamp tested — *Fixed in code: negative-grace WARN test, zero-grace immediate-refusal test, and the `>=` boundary pinned at the record level.*
- ✅ **ThroughputQuotaServiceImpl.java:354** — WARN wording branch selection unasserted — *Fixed in code: `givenDrawFailures_whenWarning_thenTheWordingNamesTheCurrentFailureDirection` asserts both wordings on their own side of the deadline; fragments match the impl strings.*
- ✅ **ThroughputQuotaServiceImpl.java:294** — second `System.nanoTime()` call in `scheduleDrawIfNeeded` — *Fixed in code: single hoisted `now` feeds both the dry-backoff check and the probe gate.*
- ✅ **ThroughputQuotaServiceImpl.java:456** — grace-trumps-window precedence re-encoded at three call sites — *Fixed in code: `RedisHealth.failingOpen(now, graceNanos)` carries the precedence and `failOpenWindowActive` is now private, so the window cannot be consulted alone; precedence pinned by a record test. All three sites verified semantically equivalent to the pre-refactor logic.*
- ✅ **ThroughputQuotaServiceImpl.java:63** — probe-interval seam's non-volatile contract undocumented — *Fixed in code: the comment now states the contract (production never writes it, tests assign on the driving thread).*
- ✅ **ThroughputQuotaServiceImplTest.java:281** — outage semantics existed only at mock level — *Fixed in code: `TotalThroughputQuotaRedisOutageIntegrationTestCase` pauses the real Valkey container, asserts refusal through the real `droppedMsgs` counter and a real 0x97 PUBACK, then asserts recovery through a real probe draw.*
- ✅ **ThroughputQuotaServiceImplTest.java:283** — five grace tests slept against real time with tight margins — *Fixed in code: `LongSupplier nanoTime` seam (same non-volatile contract as the probe seam); the grace tests advance an `AtomicLong` instead of sleeping.*

---

application/src/test/java/org/thingsboard/mqtt/broker/service/testing/integration/TotalThroughputQuotaRedisOutageIntegrationTestCase.java
```diff
@@ -0,0 +1,188 @@
...
+
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * The headline outage arc end to end, against a real Jedis failure rather than a mock: Valkey is PAUSED - the
+ * socket blackholes, so draws hang for the client timeout, the harsh detection path - the quota degrades and
```
dmytro-landiak, line 48: The suite's timing budget rests on an unstated environmental default: with `redis.standalone.useDefaultClientConfig:true` the Jedis socket timeout is the library's 2 s default, so a draw against the paused container fails well inside the 30 s awaits — but the yml's non-default `readTimeout` is 60000 ms, so anyone flipping that env-exposed flag turns every await here into a timeout with no hint why. Would it be worth pinning it in this class's `@TestPropertySource` (`redis.standalone.useDefaultClientConfig=false` plus a ~1 s `redis.standalone.readTimeout`, which would also speed up the failure detection), or at least naming the 2 s dependency here rather than the vague "the client timeout"?

application/src/test/java/org/thingsboard/mqtt/broker/service/testing/integration/TotalThroughputQuotaRedisOutageIntegrationTestCase.java
```diff
@@ -0,0 +1,188 @@
...
+    private void unpauseValkey() {
+        if (valkeyPaused) {
+            valkey.getDockerClient().unpauseContainerCmd(valkey.getContainerId()).exec();
+            valkeyPaused = false;
+            log.info("Unpaused the Valkey container");
+        }
+    }
+
+    /** Collects PUBACK reason codes off Paho's callback thread; everything else is irrelevant to this suite. */
+    private record ReasonCodeCollector(List<Integer> reasonCodes) implements MqttCallback {
```
dmytro-landiak, line 159: This is the third hand-rolled six-method v5 `MqttCallback` whose only job is collecting PUBACK reason codes — `TotalThroughputQuotaRefusalIntegrationTestCase` and `RateLimitsIntegrationTestCase` each carry an anonymous copy. Since this record is the cleanest of the three, consider hoisting it next to the shared scaffolding so the siblings can drop theirs. Same for the constant: `QUOTA_EXCEEDED = 151` is the third spelling of this value in the package (the siblings use bare `151`), and Paho already ships `MqttReturnCode.RETURN_CODE_QUOTA_EXCEEDED` — one shared reference would end the drift.

PR #366, opened 2026-09-09
Add REST API for publishing MQTT messages

roger-wang-2026:
## Pull Request description

Closes #228.

Adds a SYS_ADMIN REST endpoint for publishing MQTT messages through the standard TBMQ message pipeline:

`POST /api/mqtt/publish`

The endpoint accepts a topic, Base64-encoded payload, QoS, retained flag, MQTT 5 message expiry interval, and content type. Accepted messages go through the existing incoming throughput quota, retained-message processing, publish queue, subscription matching, and cluster downlink delivery flow.

Response semantics:
- `202 Accepted`: the message was accepted by the broker queue.
- `429 Too Many Requests`: the incoming throughput quota was exceeded.
- `503 Service Unavailable`: queue delivery failed.

The maximum decoded payload size is configurable with `MQTT_REST_PUBLISH_MAX_PAYLOAD_SIZE` and defaults to 65536 bytes.

Tests were added for the external publish service and REST controller, covering normal publishing, MQTT 5 properties, retained messages, quota rejection, queue failure, payload limits, authorization, and request validation.

Validation performed:
- Maven reactor compile for `application` and all required modules: passed.
- `git diff --check`: passed.
- Targeted tests were not completed locally because Maven began downloading the full Testcontainers/Docker test dependency stack and the dependency source stalled.

Documentation follow-up:
- Document the new endpoint, Base64 payload encoding, response codes, and `MQTT_REST_PUBLISH_MAX_PAYLOAD_SIZE` setting in the REST API documentation.

## General checklist

- [x] You have reviewed the guidelines [document](https://docs.google.com/document/d/1A4Vc3wrHsY_159b9RG5LOtCryoH6VPwgOhrFEzJKwbI/edit?usp=sharing).
- [x] [Labels](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels#about-labels) that classify your pull request have been added.
- [x] The [milestone](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/about-milestones) is specified and corresponds to fix version.
- [x] Description references specific [issue](https://github.com/thingsboard/tbmq/issues).
- [x] Description contains human-readable scope of changes.
- [x] Description contains brief notes about what needs to be added to the documentation.
- [x] No merge conflicts, commented blocks of code, code formatting issues.
- [x] Changes are backward compatible or upgrade script is provided.

## Front-End feature checklist

Not applicable; this change does not modify the UI.

## Back-End feature checklist

[...]

PR #367, opened 2026-09-09
feat(rest): REST API for publishing MQTT messages

### LINKED ISSUE #228: [Feature Request] rest api for publish messages to device   from app
rest api for publish messages to device   from app

dmytro-landiak:
## Pull Request description

Adds a REST endpoint for publishing MQTT messages through the broker, so backend services can publish without opening an MQTT connection.

Closes #228. Builds on the contribution in #366 by @roger-wang-2026 (merged into this branch).

### Endpoint

`POST /api/mqtt/publish` (SYS_ADMIN)

```json
{
  "topic": "devices/a/commands",
  "payload": "eyJjbWQiOiAicmVib290In0=",
  "payloadEncoding": "BASE64",
  "qos": 1,
  "retain": false,
  "properties": {
    "payloadFormatIndicator": 1,
    "messageExpiryInterval": 600,
    "contentType": "application/json",
    "responseTopic": "devices/a/replies",
    "correlationData": "cmVxLTQy",
    "userProperties": { "source": "backend" }
  }
}
```

- `payloadEncoding`: `BASE64` (default, raw bytes as a Base64 string), `TEXT` (string published as UTF-8), `JSON` (any JSON value published as its compact JSON text).
- MQTT 5 publish properties under `properties`: payload format indicator, message expiry interval, content type, response topic, correlation data and user properties. An empty payload with `retain` clears the retained message.
- Responses: `200` reason code 0 (the queue accepted it and at least one subscription matched the topic), `202` reason code 16 (accepted, no subscription matched). A 2xx means the broker queue accepted the message, not that any client received it. `400` invalid request, `411` missing `Content-Length` (chunked bodies are not accepted), `413` body too large, `415` Content-Type is not JSON, `429` total throughput quota exceeded, `503` publish queue failed or timed out (on timeout the message may still have been published). All documented in Swagger.

### Behaviour

- Publishes go through the same pipeline as MQTT clients: total throughput quota, retained store, publish queue, subscription matching. Publisher id is the reserved `tbmq_rest_api`.
- Refused/failed publishes count towards `droppedMsgs`; a dedicated `restPublishMsgs{result=accepted|quota_exceeded|failed}` counter is added.
- Request size is checked on `Content-Length` before deserialization (bound: 2 × `max_payload_size` + 16 KB envelope).
- Broker-wide, for every REST controller: an unreadable JSON request body (malformed JSON, unknown enum value) now answers `400` with the Jackson reason instead of `500`, and a missing or wrong `Content-Type` answers `415` instead of `500`.

### Configuration

| Env var | Default | Description |
|---|---|---|
| `REST_PUBLISH_MAX_PAYLOAD_SIZE` | `65536` | Max decoded payload size in bytes (keep below Kafka `max.request.size`) |
| `REST_PUBLISH_TIMEOUT_MS` | `30000` | Time to wait for the publish queue ack before answering 503 |

### Documentation

- New REST endpoint page: request/response shape, encodings, reason codes, required authority.
- Configuration reference: the two new env vars.
- Metrics reference: `restPublishMsgs` counter.

## General checklist

## Back-End feature checklist


application/src/main/java/org/thingsboard/mqtt/broker/controller/MqttPublishController.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @Value("${server.rest_publish.timeout_ms:30000}")
+    private long timeoutMs;
+
+    @ApiOperation(value = "Publish MQTT message (publishMqttMessage)",
+            notes = "Publishes a message to all MQTT clients whose subscriptions match the topic, through the same pipeline as a " +
+                    "PUBLISH received from a client: total throughput quota, retained-message store, publish queue and delivery. " +
+                    "The publisher client id is '" + BrokerConstants.REST_API_CLIENT_ID + "'. " +
+                    "A 2xx response means the broker queue accepted the message, not that any client received it. " +
+                    "Payloads are UTF-8 text by default; set 'payloadEncoding' to BASE64 for binary data.")
```
dmytro-landiak: This contradicts the actual default: `RestPublishRequest.payloadEncoding` is initialised to `BASE64`, and the DTO's own `@Schema` says so. Since this text lands in the generated OpenAPI docs, a caller following it will send raw text and get a "not valid Base64" 400. Looks like a leftover from before the default was flipped in `f7c8662fd` — it should read the other way round (BASE64 by default, TEXT for UTF-8 strings, JSON for JSON values).

application/src/main/java/org/thingsboard/mqtt/broker/controller/MqttPublishController.java
```diff
@@ -0,0 +1,110 @@
...
+import org.thingsboard.mqtt.broker.exception.ThingsboardErrorResponse;
+import org.thingsboard.mqtt.broker.service.mqtt.publish.RestPublishService;
+
+@Slf4j
+@RestController
+@RequestMapping("/api/mqtt")
+@RequiredArgsConstructor
+public class MqttPublishController extends BaseController {
+
+    public static final String PUBLISH_PATH = "/api/mqtt/publish";
```
dmytro-landiak, line 50: The path literal now exists four times: this constant, the `@RequestMapping("/api/mqtt")` + `@PostMapping("/publish")` pair, and a private `PUBLISH_URL` in both `MqttPublishControllerTest` and `RestPublishIntegrationTestCase`. If the mapping ever moves, the interceptor's `addPathPatterns` silently stops guarding the endpoint — a size limit failing open is a quiet failure. Building the annotations from constants and having the tests reference `MqttPublishController.PUBLISH_PATH` would make the coupling compiler-enforced.

application/src/main/java/org/thingsboard/mqtt/broker/controller/MqttPublishController.java
```diff
@@ -0,0 +1,110 @@
...
+            @ApiResponse(responseCode = "429", description = "Refused by the total incoming throughput quota.",
+                    content = @Content(schema = @Schema(implementation = ThingsboardErrorResponse.class))),
+            @ApiResponse(responseCode = "503", description = "The broker queue rejected the message or did not acknowledge it in time.",
+                    content = @Content(schema = @Schema(implementation = ThingsboardErrorResponse.class)))
+    })
+    @PreAuthorize("hasAuthority('SYS_ADMIN')")
+    @PostMapping("/publish")
+    public DeferredResult<ResponseEntity<?>> publish(@Valid @RequestBody RestPublishRequest request) {
+        DeferredResult<ResponseEntity<?>> result = new DeferredResult<>(timeoutMs);
+        result.onTimeout(() -> {
```
dmytro-landiak, line 88: When this fires the message has most likely already been handed to the Kafka producer and may still be delivered (the service will later call `stats.increment(ACCEPTED)` and `result.setResult` will just return `false`). So a 503 here is "unknown outcome", not "not published" — a caller that retries on 503 gets duplicates. Worth stating in the 503 `@ApiResponse` description (and the docs page) that the message may still have been published, so callers know to make the payload idempotent rather than retry blindly.

application/src/main/java/org/thingsboard/mqtt/broker/config/RestPublishRequestSizeInterceptor.java
```diff
@@ -0,0 +1,83 @@
...
+    @Override
+    public void addInterceptors(InterceptorRegistry registry) {
+        registry.addInterceptor(this).addPathPatterns(MqttPublishController.PUBLISH_PATH);
+    }
+
+    @Override
+    public boolean preHandle(HttpServletRequest request, HttpServletResponse response, Object handler) throws IOException {
+        long contentLength = request.getContentLengthLong();
+        if (contentLength < 0) {
+            return reject(response, HttpStatus.LENGTH_REQUIRED, "Content-Length header is required");
```
dmytro-landiak, line 73: Two things about the 411 branch. It isn't in the `@ApiResponses` list, so it's an undocumented response code. And it refuses every request sent with `Transfer-Encoding: chunked`, which several HTTP clients use by default for streamed bodies (Java `HttpClient` with `BodyPublishers.ofInputStream`, Node/undici with a stream body, some reverse proxies re-chunking). Is requiring Content-Length intentional? An alternative is to let a length-less request through to the decoded-payload check in `RestPublishServiceImpl` (the allocation bound is then Tomcat's `maxSwallowSize`/`maxPostSize`), or to wrap the input stream in a counting stream that aborts past `maxRequestBytes`. Either way the branch has no test — `MqttPublishControllerTest` covers 413 and within-limit but never a request with the header cleared.

application/src/main/java/org/thingsboard/mqtt/broker/config/RestPublishRequestSizeInterceptor.java
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * Room for the JSON envelope around the payload: topic, flags and the MQTT 5 properties block.
+     */
+    static final long ENVELOPE_ALLOWANCE_BYTES = 16 * 1024;
+
+    private final long maxRequestBytes;
+
+    public RestPublishRequestSizeInterceptor(@Value("${server.rest_publish.max_payload_size:65536}") long maxPayloadSize) {
+        // Base64 grows the payload by 4/3 (rounded up to a multiple of 4); PLAIN text never exceeds that
```
dmytro-landiak: There is no `PLAIN` encoding — the enum is BASE64/TEXT/JSON. More importantly, the claim that text "never exceeds" the Base64 bound doesn't hold for JSON-escaped text: clients that escape non-ASCII as `\uXXXX` (Python's `json.dumps` does this by default) turn a 3-byte UTF-8 character into 6 bytes, i.e. 2× growth versus Base64's 4/3. A TEXT payload of ~55 KB of non-ASCII text — within the 64 KB decoded limit — would be refused with 413 by this guard. Might be worth sizing the bound for the worst realistic encoding (2×, or a configurable multiplier) or at least noting the limitation in the yml comment.

application/src/main/java/org/thingsboard/mqtt/broker/config/RestPublishRequestSizeInterceptor.java
```diff
@@ -0,0 +1,83 @@
...
+    @Override
+    public boolean preHandle(HttpServletRequest request, HttpServletResponse response, Object handler) throws IOException {
+        long contentLength = request.getContentLengthLong();
+        if (contentLength < 0) {
+            return reject(response, HttpStatus.LENGTH_REQUIRED, "Content-Length header is required");
+        }
+        if (contentLength > maxRequestBytes) {
+            log.debug("Refusing REST publish request of {} bytes, limit is {} bytes", contentLength, maxRequestBytes);
+            return reject(response, HttpStatus.PAYLOAD_TOO_LARGE,
+                    "Request body of " + contentLength + " bytes exceeds the maximum of " + maxRequestBytes + " bytes");
```
dmytro-landiak: The number reported here is the derived request-body budget (`payload*4/3 + 16 KB`), not the `max_payload_size` the operator configured, so a user who set 65536 sees a 413 quoting ~103 KB and can't tell what payload size is actually allowed. Including the configured payload limit in the message would make it actionable. Relatedly, the 16 KB envelope allowance isn't tied to any validated bound on the properties block (`userProperties` is an unbounded map), so a large properties block can push a within-limit payload past the guard, or let an over-limit payload through to the later check — the two limits are hard to reason about together.

application/src/main/java/org/thingsboard/mqtt/broker/config/RestPublishRequestSizeInterceptor.java
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Refuses oversized REST publish requests before the body is read. The decoded-payload check in
+ * {@link org.thingsboard.mqtt.broker.service.mqtt.publish.RestPublishServiceImpl} runs only after Jackson has
+ * materialised the whole JSON body, so without this guard {@code server.rest_publish.max_payload_size} would bound
+ * what gets published but not what gets allocated.
+ */
+@Slf4j
+@Component
+public class RestPublishRequestSizeInterceptor implements HandlerInterceptor, WebMvcConfigurer {
```
dmytro-landiak, line 43: One class playing both `HandlerInterceptor` and `WebMvcConfigurer` means a global MVC configuration hook lives on the interceptor itself; any future `WebMvcConfigurer` callback added here applies broker-wide from a class named "...RequestSizeInterceptor". The usual split — a plain interceptor plus a small `@Configuration implements WebMvcConfigurer` that registers it — keeps the seam obvious and makes the interceptor unit-testable without a Spring context. Minor, since this is the first interceptor in the codebase.

dmytro-landiak (commented):
## Review summary

Reviewed 28 changed files in [feat(rest): REST API for publishing MQTT messages](https://github.com/thingsboard/tbmq/pull/367). Left 16 comment(s) inline.

The pipeline mirroring is sound: quota is charged before retained processing (same order as `MqttSessionHandler` → `MqttPublishHandler`), the queue send reuses `sendToPublishQueue`, `TbRateLimitsException` and `DataValidationException` reach `ThingsboardErrorResponseHandler` as 429/400, and the noLocal / subscription-matching path handles the session-less `tbmq_rest_api` sender without a client-session lookup. No functional bugs found. The inline comments are about contract/documentation mismatches (the Swagger note contradicts the BASE64 default, an undocumented 411 and an undocumented `success` response field), the Content-Length guard's edge cases, and a handful of maintainability points.

---

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/publish/RestPublishServiceImpl.java
```diff
@@ -0,0 +1,206 @@
...
+            tbMessageStatsReportClient.reportDroppedMsgs();
+            stats.increment(RestPublishOutcome.QUOTA_EXCEEDED);
+            throw new TbRateLimitsException("Total message rate limit exceeded");
+        }
+
+        if (publishMsg.isRetained()) {
+            publishMsg = retainedMsgProcessor.process(publishMsg);
+        }
+
+        boolean hasSubscribers = !subscriptionService.getSubscriptions(publishMsg.getTopicName()).isEmpty();
```
dmytro-landiak, line 94: This runs a full trie lookup purely to pick 200 vs 202, and `MsgDispatcherServiceImpl.processPublishMsg` repeats the exact same `getSubscriptions(topic)` when the message comes off the queue — so every REST publish pays the wildcard walk twice and feeds `subscriptionsLookup` timer stats a second sample. It's also only approximate: the list includes subscriptions whose client session is already gone (`convertToSubscription` returns null for those later) and subscriptions that noLocal will exclude, so a 200 can be reported for zero actual deliveries; conversely a subscriber connecting between this check and consumption gets the message despite a 202. Given that, is the pre-check worth it? Options: a cheap `hasSubscriptions(topic)` short-circuit on `SubscriptionService`, or always answering 202 with a documented "queued" semantic and dropping reason code 16 from the contract.

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/publish/RestPublishServiceImpl.java
```diff
@@ -0,0 +1,206 @@
...
+
+    @Override
+    public ListenableFuture<RestPublishResponse> publish(RestPublishRequest request) {
+        PublishMsg publishMsg = toPublishMsg(request);
+
+        if (!throughputQuotaService.tryConsumeIncoming()) {
+            log.warn("[{}] REST publish refused by the total throughput quota, payload size: {}", publishMsg.getTopicName(), publishMsg.getPayload().length);
+            tbMessageStatsReportClient.reportDroppedMsgs();
+            stats.increment(RestPublishOutcome.QUOTA_EXCEEDED);
+            throw new TbRateLimitsException("Total message rate limit exceeded");
```
dmytro-landiak, line 87: `ThingsboardErrorResponseHandler#handleRateLimitException` hardcodes `"Too many requests!"` and ignores the exception message, so this text never reaches the caller — the 429 body says nothing about the throughput quota, which is why the controller test can only assert `$.message` exists. Either drop the bespoke message, or (more useful for an operator debugging a 429) pass it through in the handler so the response distinguishes the total-throughput quota from other rate limits.

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/publish/RestPublishServiceImpl.java
```diff
@@ -0,0 +1,206 @@
...
+    /**
+     * Folds the legacy top-level {@code messageExpiryInterval} / {@code contentType} into the properties block; the nested values win.
+     */
+    @SuppressWarnings("deprecation")
+    private static RestPublishProperties effectiveProperties(RestPublishRequest request) {
+        RestPublishProperties props = request.getProperties();
+        if (request.getMessageExpiryInterval() == null && request.getContentType() == null) {
+            return props;
+        }
+        RestPublishProperties merged = props == null ? new RestPublishProperties() : props;
```
dmytro-landiak: Despite the name, `merged` is the caller's own `request.getProperties()` instance, and the two `setX` calls below mutate the request DTO in place. It works because the request is per-call, but a static "compute the effective view" helper rewriting its input is surprising, and any later re-read of the request sees different data than the client sent. Copying into a fresh `RestPublishProperties` keeps the method as pure as its name suggests.

application/src/main/java/org/thingsboard/mqtt/broker/dto/RestPublishResponse.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public static RestPublishResponse success() {
+        return new RestPublishResponse(MqttReasonCodes.PubAck.SUCCESS.byteValue(), "Success");
+    }
+
+    public static RestPublishResponse noMatchingSubscribers() {
+        return new RestPublishResponse(MqttReasonCodes.PubAck.NO_MATCHING_SUBSCRIBERS.byteValue(), "No matching subscribers");
+    }
+
+    public boolean isSuccess() {
```
dmytro-landiak, line 45: Jackson treats `isSuccess()` as a bean getter, so every response body carries a third field `"success": true|false` that the class-level `@Schema`, the `@ApiResponse` docs and the PR description don't mention — the published contract and the actual body disagree, and the field is redundant with `reasonCode == 0`. It also has no setter, so the DTO doesn't round-trip through a strict ObjectMapper. Since it's only used by the controller to pick the status code, `@JsonIgnore` (or documenting it deliberately) would close the gap.

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/publish/RestPublishService.java
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * Publishes an MQTT message on behalf of the REST API through the standard broker pipeline: total throughput
+     * quota, retained-message processing, publish queue, subscription matching and downlink.
+     * <p>
+     * Request-level problems are reported synchronously: {@link org.thingsboard.mqtt.broker.exception.DataValidationException}
+     * for an invalid topic, payload or property, {@link org.thingsboard.mqtt.broker.exception.TbRateLimitsException}
+     * when the total throughput quota refuses the message. The returned future completes once the publish queue
+     * acknowledges the message and fails with the queue error otherwise.
+     */
+    ListenableFuture<RestPublishResponse> publish(RestPublishRequest request);
```
dmytro-landiak, line 35: The failure contract is split: invalid topic/payload/property and quota refusal are thrown synchronously, while queue rejection comes back as a failed future. Any caller other than this controller has to write two error paths for one call, and it's easy to forget the try/catch around what looks like a non-blocking API. If the sync throw is deliberate (it's what routes `TbRateLimitsException` to the global handler for the 429 mapping), that trade-off is worth stating in the javadoc rather than just listing the exception types; otherwise returning `Futures.immediateFailedFuture(...)` for those cases gives one failure channel.

application/src/main/java/org/thingsboard/mqtt/broker/controller/BaseController.java
```diff
@@ -190,6 +193,19 @@ private ThingsboardException handleException(Exception exception, boolean logExc
...
+    /**
+     * Jackson appends the parser location and the reference chain after the first line break; only the sentence
+     * before it explains what was wrong with the body.
+     */
+    private static String unreadableBodyReason(HttpMessageNotReadableException e) {
+        String message = e.getMostSpecificCause().getMessage();
+        if (message == null) {
+            return e.getMessage();
+        }
+        int cut = message.indexOf('\n');
```
dmytro-landiak: Two thoughts. This parses Jackson's exception text by looking for a newline — an undocumented formatting detail that has changed across Jackson versions and doesn't hold for every `JsonProcessingException` subtype, so the trimming can silently stop working and start leaking `[Source: ...]` again; the existing test only asserts the negative on one message shape. `JsonMappingException#getOriginalMessage()` gives the same thing without string surgery (and the `message == null` fallback is untested). Second, since this lives in `BaseController` it changes the response for *every* controller — unreadable bodies used to fall through to GENERAL/500 and now return 400. That's an improvement, but it's a broker-wide behaviour change that the PR description doesn't mention.

application/src/main/java/org/thingsboard/mqtt/broker/service/stats/StatsManagerImpl.java
```diff
@@ -527,6 +535,12 @@ public void printStats() {
         log.info("[{}] Stats: count = [{}]", StatsType.THROUGHPUT_QUOTA_DEGRADED.getPrintName(), throughputQuotaStats.getCount());
         throughputQuotaStats.reset();
 
+        log.info("[{}] Stats: accepted = [{}] quotaExceeded = [{}] failed = [{}]", StatsType.REST_PUBLISH_MSGS.getPrintName(),
```
dmytro-landiak: The three outcomes are spelled out by hand here, so adding a fourth `RestPublishOutcome` compiles fine and just silently disappears from the periodic log. Iterating `RestPublishOutcome.values()` into a joined `name = [count]` string would keep it self-maintaining.

common/data/src/main/java/org/thingsboard/mqtt/broker/common/data/BrokerConstants.java
```diff
@@ -200,6 +200,11 @@ public static String getQosSentStatsKey(int qos) {
     public static final String WEB_SOCKET_DEFAULT_SUBSCRIPTION_COLOR = "#34920C";
 
     public static final String SYSTEM_DUMMY_CLIENT_ID_PREFIX = "tbmq_system_dummy_client_id_";
+    /**
+     * Client id stamped on messages published through the REST API ({@code POST /api/mqtt/publish}). Reserved:
+     * it drives the {@code noLocal} check and appears as the sender in client logs, so no real client should use it.
+     */
+    public static final String REST_API_CLIENT_ID = "tbmq_rest_api";
```
dmytro-landiak, line 207: "Reserved" here is documentation only — nothing stops a real MQTT client from connecting as `tbmq_rest_api` (unlike `SYSTEM_DUMMY_CLIENT_ID_PREFIX` there's no check anywhere). If one does, its noLocal subscriptions silently drop every REST publish and its client-log stream gets interleaved with REST traffic. Should CONNECT refuse this client id (CLIENT_IDENTIFIER_NOT_VALID), or is best-effort documentation enough here?

application/src/test/java/org/thingsboard/mqtt/broker/service/testing/integration/TotalThroughputQuotaRestPublishIntegrationTestCase.java
```diff
@@ -0,0 +1,101 @@
...
+        // one drop for the refusal, plus one per accepted publish: nobody subscribes to the topic, so the publish
+        // consumer counts each of them as dropped too, asynchronously
+        awaitDroppedMsgsAtLeast("refusal and no-subscriber drops counted", droppedBefore + 1 + accepted);
+        assertThat(droppedMsgs()).isEqualTo(droppedBefore + 1 + accepted);
+        assertThat(restPublishCount(RestPublishOutcome.QUOTA_EXCEEDED)).isEqualTo(quotaExceededBefore + 1);
+        assertThat(restPublishCount(RestPublishOutcome.ACCEPTED)).isEqualTo(acceptedBefore + accepted);
+    }
+
+    private double restPublishCount(RestPublishOutcome outcome) {
+        return meterRegistry.get(StatsType.REST_PUBLISH_MSGS.getPrintName()).tag("result", outcome.getTagValue()).counter().count();
```
dmytro-landiak: `"result"` duplicates `DefaultRestPublishStats.RESULT_TAG` (which `DefaultRestPublishStatsTest` already references). Making the constant public, or importing it here, keeps the tag name in one place so a rename doesn't leave this test looking up a meter that no longer exists.

application/src/main/java/org/thingsboard/mqtt/broker/controller/MqttPublishController.java
dmytro-landiak: Fixed in 73a09d330: the note now says the payload is a Base64 string by default, with TEXT for a UTF-8 string and JSON for any JSON value.

application/src/main/java/org/thingsboard/mqtt/broker/controller/MqttPublishController.java
dmytro-landiak: Fixed in 73a09d330: the class-level @RequestMapping is gone, @PostMapping uses PUBLISH_PATH, and both tests reference MqttPublishController.PUBLISH_PATH.

application/src/main/java/org/thingsboard/mqtt/broker/controller/MqttPublishController.java
dmytro-landiak: Fixed in 73a09d330: the 503 description states that on a timeout the message may still have been published and retries can produce duplicates. The PR description carries the same note for the docs page.

application/src/main/java/org/thingsboard/mqtt/broker/config/RestPublishRequestSizeInterceptor.java
dmytro-landiak: Partially fixed in 73a09d330: 411 is now in @ApiResponses and covered by a test (request without Content-Length). Requiring Content-Length stays intentional: accepting chunked bodies would need a filter with a counting stream and a 413 mapping out of HttpMessageNotReadableException, which is not worth it for a small JSON body. Clients that stream the body can buffer it (e.g. BodyPublishers.ofString / ofByteArray).

application/src/main/java/org/thingsboard/mqtt/broker/config/RestPublishRequestSizeInterceptor.java
dmytro-landiak: Fixed in 73a09d330: the stale PLAIN comment is gone and the bound is now 2 x max_payload_size + 16 KB, covering \uXXXX-escaped text as well as Base64. The yml comment describes the formula.

application/src/main/java/org/thingsboard/mqtt/broker/config/RestPublishRequestSizeInterceptor.java
dmytro-landiak: Fixed in 73a09d330: the 413 message now quotes the configured max_payload_size alongside the derived body limit. The envelope allowance stays a fixed 16 KB; it only bounds allocation, the payload itself is checked exactly after decoding.

application/src/main/java/org/thingsboard/mqtt/broker/config/RestPublishRequestSizeInterceptor.java
dmytro-landiak: Declined: this is the only interceptor in the codebase and the WebMvcConfigurer callback registers just itself. Splitting into two classes adds a file without changing behaviour; can revisit if a second interceptor shows up.

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/publish/RestPublishServiceImpl.java
dmytro-landiak: Declined: the extra trie lookup costs microseconds per HTTP request, negligible next to the HTTP round trip, and reason code 16 is a deliberate part of the contract. The response description already states that a 2xx does not confirm delivery to any client, which covers the approximation.

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/publish/RestPublishServiceImpl.java
dmytro-landiak: Fixed in 73a09d330: handleRateLimitException now passes the exception message through (falling back to the old text when it is null), so the 429 body says "Total message rate limit exceeded" and the controller test asserts it. This is the only TbRateLimitsException thrown in main code, so no other 429 body changes.

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/publish/RestPublishServiceImpl.java
dmytro-landiak: Fixed in 73a09d330: effectiveProperties copies into a fresh RestPublishProperties and leaves the request untouched; a test asserts the request is not mutated.

application/src/main/java/org/thingsboard/mqtt/broker/dto/RestPublishResponse.java
dmytro-landiak: Fixed in 73a09d330: isSuccess() is @JsonIgnore, and the 200 test asserts the field is absent from the body.

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/publish/RestPublishService.java
dmytro-landiak: Fixed in 73a09d330: the javadoc now explains the trade-off, the sync throw is deliberate because the message is refused before anything is queued and the global handler maps it to 400 / 429.

application/src/main/java/org/thingsboard/mqtt/broker/controller/BaseController.java
dmytro-landiak: Fixed in 73a09d330: the reason now comes from JsonProcessingException.getOriginalMessage() when the cause is a Jackson exception, with the plain cause message as fallback. The broker-wide 500 -> 400 change is now called out in the PR description.

application/src/main/java/org/thingsboard/mqtt/broker/service/stats/StatsManagerImpl.java
dmytro-landiak: Fixed in 73a09d330: the log line iterates RestPublishOutcome.values().

common/data/src/main/java/org/thingsboard/mqtt/broker/common/data/BrokerConstants.java
dmytro-landiak: Declined for now: SYSTEM_DUMMY_CLIENT_ID_PREFIX is not enforced on CONNECT either, so this keeps the same documentation-only approach for consistency. Refusing the id at CONNECT is a small follow-up if we decide to enforce reserved ids.

application/src/test/java/org/thingsboard/mqtt/broker/service/testing/integration/TotalThroughputQuotaRestPublishIntegrationTestCase.java
dmytro-landiak: Fixed in 73a09d330: RESULT_TAG is public and the test references it.

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/publish/RestPublishServiceImpl.java
```diff
@@ -0,0 +1,214 @@
...
+     */
+    @SuppressWarnings("deprecation")
+    private static RestPublishProperties effectiveProperties(RestPublishRequest request) {
+        RestPublishProperties props = request.getProperties();
+        if (request.getMessageExpiryInterval() == null && request.getContentType() == null) {
+            return props;
+        }
+        RestPublishProperties merged = new RestPublishProperties();
+        if (props != null) {
+            merged.setPayloadFormatIndicator(props.getPayloadFormatIndicator());
```
dmytro-landiak: The non-mutating copy lists all six fields by hand, so the next property added to `RestPublishProperties` is silently dropped from the merged view and the compiler won't say anything. Since the DTO is already `@Data`, adding `@Builder(toBuilder = true)` + `@AllArgsConstructor`/`@NoArgsConstructor` (or a small copy constructor on the DTO) would keep the field list next to the fields themselves.

application/src/test/java/org/thingsboard/mqtt/broker/controller/MqttPublishControllerTest.java
```diff
@@ -0,0 +1,307 @@
...
+                .andExpect(status().isBadRequest())
+                .andExpect(jsonPath("$.message").value(org.hamcrest.Matchers.allOf(
+                        org.hamcrest.Matchers.startsWith("Invalid request body:"),
+                        org.hamcrest.Matchers.not(org.hamcrest.Matchers.containsString("[Source:")))));
+        verify(restPublishService, never()).publish(any());
+    }
+
+    @Test
+    public void givenBodyLargerThanConfiguredLimit_whenPublish_thenPayloadTooLargeBeforeDeserialization() throws Exception {
+        // 1024-byte payload cap -> ~1.4 KB base64 + envelope allowance; 64 KB of body must be refused up front
```
dmytro-landiak: This comment still describes the old `4/3` Base64 derivation, while the interceptor now uses `maxPayloadSize * 2 + 16 KB`. Neither this test nor `givenBodyWithinLimit_...` exercises the boundary the fix changed — a body between the old 4/3 bound and the new 2× bound (e.g. ~1.9 KB with a 1024 cap) should now pass the guard, and one byte over `2*1024 + 16K` should still be refused. Without a boundary case the factor can regress to any value unnoticed.

application/src/test/java/org/thingsboard/mqtt/broker/controller/MqttPublishControllerTest.java
```diff
@@ -0,0 +1,307 @@
...
+        RestPublishRequest request = validRequest();
+        request.setPayload(new TextNode("x".repeat(64 * 1024)));
+
+        doPost(PUBLISH_URL, request)
+                .andExpect(status().isPayloadTooLarge())
+                .andExpect(jsonPath("$.message").value(org.hamcrest.Matchers.containsString("max_payload_size is 1024 bytes")));
```
dmytro-landiak: Minor: this is the fourth fully-qualified `org.hamcrest.Matchers.*` reference in the class — a static import would read cleaner. Also the sibling 411 test asserts `$.errorCode` is 31 while this 413 test doesn't; worth making the two size-guard tests symmetric.

application/src/main/java/org/thingsboard/mqtt/broker/controller/BaseController.java
```diff
@@ -190,6 +194,18 @@ private ThingsboardException handleException(Exception exception, boolean logExc
         return new ThingsboardException(exception.getMessage(), exception, ThingsboardErrorCode.GENERAL);
     }
 
+    /**
+     * Jackson's full message carries the parser location and the reference chain; only the original message
+     * explains what was wrong with the body.
+     */
+    private static String unreadableBodyReason(HttpMessageNotReadableException e) {
+        Throwable cause = e.getMostSpecificCause();
+        if (cause instanceof JsonProcessingException jpe) {
```
dmytro-landiak, line 203: The non-Jackson fallback branch below (`cause.getMessage()` / `e.getMessage()`) has no test, and `givenUnknownPayloadEncoding_...` only asserts the negative (`[Source:` absent). A positive assertion on the expected reason text (e.g. that it mentions `HEX` or the accepted enum values) would pin down the `getOriginalMessage()` behaviour rather than just the absence of the old one.

dmytro-landiak (commented):
## Re-review summary

Re-reviewed [feat(rest): REST API for publishing MQTT messages](https://github.com/thingsboard/tbmq/pull/367) — verified 16 finding(s) from [previous review](https://github.com/thingsboard/tbmq/pull/367#pullrequestreview-5164060940).

| Status | Count |
|--------|-------|
| ✅ Resolved | 13 |
| 💬 Acknowledged | 3 |

All previous findings have been addressed. Also left 4 minor comment(s) inline on the fix commit `73a09d330` (test coverage / maintainability, no functional issues found).

### Finding details

- ✅ **MqttPublishController.java** — Swagger text claimed UTF-8 text is the default encoding — *Fixed in code (now describes BASE64 default, TEXT and JSON).*
- ✅ **MqttPublishController.java** — publish path literal duplicated four times — *Fixed in code (`@PostMapping(PUBLISH_PATH)`, tests reference the constant).*
- ✅ **MqttPublishController.java** — 503 timeout means unknown outcome, retries may duplicate — *Fixed in code (503 description states it).*
- 💬 **RestPublishRequestSizeInterceptor.java** — 411 undocumented and refuses chunked bodies — *411 now in `@ApiResponses` + tested; Developer: "Requiring Content-Length stays intentional ... clients that stream the body can buffer it."*
- ✅ **RestPublishRequestSizeInterceptor.java** — stale `PLAIN` comment, 4/3 bound too tight for `\uXXXX`-escaped text — *Fixed in code (2× factor, comment updated).*
- ✅ **RestPublishRequestSizeInterceptor.java** — 413 message quoted derived bound, not `max_payload_size` — *Fixed in code.*
- 💬 **RestPublishRequestSizeInterceptor.java** — interceptor also acts as `WebMvcConfigurer` — *Developer: "only interceptor in the codebase ... can revisit if a second interceptor shows up."*
- 💬 **RestPublishServiceImpl.java** — duplicate trie lookup for 200 vs 202 is approximate — *Developer: "costs microseconds ... reason code 16 is a deliberate part of the contract."*
- ✅ **RestPublishServiceImpl.java** — 429 body ignored the exception message — *Fixed in code (`handleRateLimitException` passes the message through).*
- ✅ **RestPublishServiceImpl.java** — `effectiveProperties` mutated the request DTO — *Fixed in code (copies into a fresh instance, test added).*
- ✅ **RestPublishResponse.java** — undocumented `success` field serialized — *Fixed in code (`@JsonIgnore`, test asserts absence).*
- ✅ **RestPublishService.java** — split sync/async failure contract undocumented — *Fixed in code (javadoc explains the trade-off).*
- ✅ **BaseController.java** — newline-splitting of Jackson message, broker-wide 500→400 unmentioned — *Fixed in code (`getOriginalMessage()`), PR description updated.*
- ✅ **StatsManagerImpl.java** — outcomes spelled out by hand in the stats log — *Fixed in code (iterates `RestPublishOutcome.values()`).*
- 💬 **BrokerConstants.java** — `tbmq_rest_api` reserved only by documentation — *Developer: "SYSTEM_DUMMY_CLIENT_ID_PREFIX is not enforced on CONNECT either ... small follow-up if we decide to enforce reserved ids."*
- ✅ **TotalThroughputQuotaRestPublishIntegrationTestCase.java** — `"result"` tag literal duplicated — *Fixed in code (`RESULT_TAG` public).*

---

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/publish/RestPublishServiceImpl.java
dmytro-landiak: Fixed in 2a8dec9a9: RestPublishProperties is now @Builder(toBuilder = true) (+ no-args/all-args constructors, Jackson binding unchanged) and effectiveProperties copies via props.toBuilder().build(), so a new property cannot be dropped from the merged view.

application/src/test/java/org/thingsboard/mqtt/broker/controller/MqttPublishControllerTest.java
dmytro-landiak: Fixed in 2a8dec9a9: the stale 4/3 comment is updated, and two boundary tests pin the factor: a body of exactly 2 x max_payload_size + 16 KB passes the guard and reaches the service, one byte more is refused with 413. The test derives the bound from the interceptor's PAYLOAD_ENCODING_FACTOR / ENVELOPE_ALLOWANCE_BYTES constants (now public).

application/src/test/java/org/thingsboard/mqtt/broker/controller/MqttPublishControllerTest.java
dmytro-landiak: Fixed in 2a8dec9a9: hamcrest matchers are statically imported, and the 413 test asserts errorCode 31 like the 411 test.

application/src/main/java/org/thingsboard/mqtt/broker/controller/BaseController.java
dmytro-landiak: Partially fixed in 2a8dec9a9: the unknown-encoding test now asserts the reason text positively (contains the offending "HEX" value and Jackson's 'not one of the values accepted for Enum class' phrase) alongside the [Source: absence. The non-Jackson fallback branch is left untested: HttpMessageNotReadableException from the JSON converter always carries a Jackson cause, so the branch is only reachable with a non-Jackson converter, which the broker does not register.

application/src/test/java/org/thingsboard/mqtt/broker/controller/MqttPublishControllerTest.java
```diff
@@ -0,0 +1,327 @@
...
+
+@DaoSqlTest
+@TestPropertySource(properties = {
+        "server.rest_publish.max_payload_size=1024",
+        "server.rest_publish.timeout_ms=500"
+})
+public class MqttPublishControllerTest extends AbstractControllerTest {
+
+    private static final String PUBLISH_URL = MqttPublishController.PUBLISH_PATH;
+    private static final long MAX_PAYLOAD_SIZE = 1024;
```
dmytro-landiak: This `1024` duplicates the literal in `@TestPropertySource("server.rest_publish.max_payload_size=1024")` a few lines up, and the two can drift silently — the boundary tests would then pin the wrong bound. Could the property string be built from the constant (`"server.rest_publish.max_payload_size=" + MAX_PAYLOAD_SIZE` is a valid constant expression for the annotation), or the interceptor's bound injected via `@Value` instead of recomputed here?

application/src/test/java/org/thingsboard/mqtt/broker/controller/MqttPublishControllerTest.java
```diff
@@ -0,0 +1,327 @@
...
+
+        doPost(PUBLISH_URL, request)
+                .andExpect(status().isPayloadTooLarge())
+                .andExpect(jsonPath("$.errorCode").value(31))
+                .andExpect(jsonPath("$.message").value(containsString("max_payload_size is 1024 bytes")));
+        verify(restPublishService, never()).publish(any());
+    }
+
+    @Test
+    public void givenBodyOneByteOverLimit_whenPublish_thenPayloadTooLarge() throws Exception {
```
dmytro-landiak: Now that the +1 boundary test exists, the 64 KB `givenBodyLargerThanConfiguredLimit_...` test covers a strict subset of the same rejection path; the only thing it adds is the `errorCode`/message assertions. Consider folding those assertions into this boundary test and dropping the 64 KB one, so there's a single rejection test per behaviour rather than two overlapping ones.

application/src/test/java/org/thingsboard/mqtt/broker/controller/MqttPublishControllerTest.java
```diff
@@ -0,0 +1,327 @@
...
+import org.mockito.ArgumentCaptor;
+import org.springframework.http.MediaType;
+import org.springframework.mock.web.MockAsyncContext;
+import org.springframework.test.context.TestPropertySource;
+import org.springframework.test.context.bean.override.mockito.MockitoBean;
+import org.springframework.test.web.servlet.MvcResult;
+import org.springframework.test.web.servlet.ResultActions;
+import org.springframework.test.web.servlet.request.MockHttpServletRequestBuilder;
+import org.thingsboard.mqtt.broker.dao.DaoSqlTest;
+import org.thingsboard.mqtt.broker.config.RestPublishRequestSizeInterceptor;
```
dmytro-landiak, line 33: Nit: this import landed after `dao.DaoSqlTest`, breaking the alphabetical order the rest of the block follows (`config` sorts before `dao`).

dmytro-landiak (commented):
## Re-review summary

Re-reviewed [feat(rest): REST API for publishing MQTT messages](https://github.com/thingsboard/tbmq/pull/367) — verified 4 finding(s) from [previous review](https://github.com/thingsboard/tbmq/pull/367#pullrequestreview-5164191864).

| Status | Count |
|--------|-------|
| ✅ Resolved | 4 |
| 💬 Acknowledged | 0 |

All previous findings have been addressed. The fix commits (`2a8dec9a9`, `77a910fe4`) introduce no correctness issues; left 3 minor test-hygiene comments inline.

### Finding details

- ✅ **application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/publish/RestPublishServiceImpl.java** — hand-copied field list in `effectiveProperties` would silently drop new properties — *Fixed in code: the legacy top-level aliases and the merge helper were removed outright in `77a910fe4`; `toMqttProperties` now reads `request.getProperties()` directly.*
- ✅ **application/src/test/java/org/thingsboard/mqtt/broker/controller/MqttPublishControllerTest.java** — stale `4/3` comment and no boundary test for the `2 × max_payload_size + 16 KB` guard — *Fixed in code: comment updated; `givenBodyExactlyAtLimit_...` and `givenBodyOneByteOverLimit_...` pin the bound, derived from the now-public interceptor constants.*
- ✅ **application/src/test/java/org/thingsboard/mqtt/broker/controller/MqttPublishControllerTest.java** — fully-qualified hamcrest matchers; 413 test missing the `errorCode` assertion — *Fixed in code: static imports; 413 test asserts `errorCode` 31 like the 411 test.*
- ✅ **application/src/main/java/org/thingsboard/mqtt/broker/controller/BaseController.java** — unknown-encoding test only asserted the negative; non-Jackson fallback untested — *Fixed in code: the test now asserts the offending `"HEX"` value and Jackson's enum phrase. The non-Jackson fallback branch stays untested; the developer's point that the JSON converter always carries a Jackson cause is valid, so no further action needed.*

---

application/src/test/java/org/thingsboard/mqtt/broker/controller/MqttPublishControllerTest.java
dmytro-landiak: Fixed in 2427602f9: @TestPropertySource builds the property from MAX_PAYLOAD_SIZE, so the bound and the configured limit come from one constant.

application/src/test/java/org/thingsboard/mqtt/broker/controller/MqttPublishControllerTest.java
dmytro-landiak: Fixed in 2427602f9: the 64 KB test is gone; the +1 boundary test carries the errorCode and message assertions.

application/src/test/java/org/thingsboard/mqtt/broker/controller/MqttPublishControllerTest.java
dmytro-landiak: Fixed in 2427602f9: import order restored.

PR #373, opened 2026-09-21
fix(session): discard expired session state on reconnect

### LINKED ISSUE #372: A Session is still resumed long after the Session Expiry Interval the server granted has elapsed
**Describe the bug**

Hello, our team @AIS2Lab found a Session lifetime conformance issue in TBMQ while testing MQTT v5
session state across connection boundaries.

TBMQ grants a Session Expiry Interval of 3 seconds in CONNACK and then keeps the Session
indefinitely. A `Clean Start 0` reconnect 16 seconds later still answers `Session Present = 1`, the
subscription made before the disconnect is still active, and a message published after the reconnect
is delivered through it.

Section 3.1.2.11.2 of MQTT v5.0 ends a Session that many seconds after the Network Connection is
closed, and section 3.2.2.1.1 requires `Session Present` to be 0 when the Server holds no Session
State for the Client Identifier. The interval here is the one **the server itself granted**: the
CONNACK carries the Session Expiry Interval property with the value 3, so this is not a case of the
client asking for something the server declined.

**Your Server Environment**

* Deployment: local `docker compose`, official `thingsboard/tbmq:2.4.0` image
  (digest `sha256:56d296f0df2853ee87e433e8cd209e721dc93615d33bfb08b5354c0063da3475`) with
  PostgreSQL 17, Kafka 4.0.0 and Valkey 8.0, single node, plain TCP listener on 1883,
  anonymous access (the default authentication providers are disabled)
* TBMQ version: 2.4.0 (the container logs report `v2.4.0`)
* OS name and version: Ubuntu 22.04.5 LTS, kernel 5.15.0-191-generic, x86_64;
  Docker Engine 29.4.3 with Compose 5.1.3. The TBMQ container itself is Debian 13 (trixie).

**Your Client Environment**

**Your Device**

* Connectivity: **MQTT**
* Device vendor and model: none. The client is a Python 3 script using raw sockets and hand-written
  MQTT v5.0 framing, so that no client library sits between the observation and the wire. It is
  attached at the end of this report.

**To Reproduce**

1. `CONNECT` with `Clean Start = 1` and Session Expiry Interval 3. Note the Session Expiry Interval
   the server returns in CONNACK.
2. `SUBSCRIBE` to a topic at QoS 1.
3. `DISCONNECT` with Reason Code 0x00.
4. Wait 16 seconds, more than five times the granted interval.
5. `CONNECT` again with the same Client Identifier and `Clean Start = 0`.
6. From a third client, publish to the topic from step 2.

Wire trace of exactly that, against TBMQ 2.4.0:

```
--> CONNECT   clean_start=1, Session Expiry Interval = 3
<-- CONNACK   2018000015110000000322000013003c2103e827000100002901
              ^^ session_present=0
                    ^^^^^^^^^^^^ property 0x11 = 00 00 00 03, the server grants 3 seconds
--> SUBSCRIBE
<-- SUBACK    900400010001
--> DISCONNECT 0x00                     e0020000

    ... 16 seconds pass, the Session should have ended after 3 ...

--> CONNECT   clean_start=0, same Client Identifier
<-- CONNACK   2018010015110000000322000013003c2103e827000100002901
              ^^ session_present=1   <-- expected 0

--> (third client) PUBLISH  xcv/.../a  "after-expiry"
<-- PUBLISH   'xcv/.../a' qos=1 pid=1 payload=b'after-expiry'
              the subscription created before the disconnect is still delivering
```

**Expected behavior**

At step 5 the CONNACK should carry `Session Present = 0`, and the subscription from step 2 should be
gone, so that step 6 delivers nothing.

**Additional context**

*Control.* The script runs a control in the same execution: a second identity with the identical
setup reconnects after **1 second**, inside the granted interval, and correctly gets
`Session Present = 1`. That matters, because it proves the broker really does create a Session with
this interval and that the test is able to observe a correct `Session Present = 0`. Only the
reconnect after the interval has elapsed behaves unexpectedly.

*Reproduction rate.* 3 of 3 runs, each with a freshly generated Client Identifier and topic
namespace, on a freshly started deployment.

*Reproducer.* Standard library only, no dependencies. Exit code 1 means the behaviour was
reproduced. It prints every frame it sends and receives, so its output can be pasted back into this
thread as-is.

```
python3 repro.py <host> 1883
```

<details>
<summary>repro.py</summary>

```python
#!/usr/bin/env python3
"""Self-contained MQTT v5 reproducer. Standard library only, no dependencies.

Usage:  python3 <script>.py [host] [port]
"""
import socket, struct, sys, time, uuid

SETTLE = 1.0  # seconds to let a new subscription take effect before it is used

CONNECT, CONNACK, PUBLISH, PUBACK, PUBREC, PUBREL, PUBCOMP = 1, 2, 3, 4, 5, 6, 7
SUBSCRIBE, SUBACK, UNSUBSCRIBE, UNSUBACK, PINGREQ, PINGRESP, DISCONNECT = 8, 9, 10, 11, 12, 13, 14
NAMES = {1:"CONNECT",2:"CONNACK",3:"PUBLISH",4:"PUBACK",5:"PUBREC",6:"PUBREL",7:"PUBCOMP",
         8:"SUBSCRIBE",9:"SUBACK",10:"UNSUBSCRIBE",11:"UNSUBACK",12:"PINGREQ",13:"PINGRESP",14:"DISCONNECT"}
P_MSG_EXPIRY, P_SESSION_EXPIRY, P_WILL_DELAY, P_RECV_MAX, P_MAX_PKT = 0x02, 0x11, 0x18, 0x21, 0x27
P_SUB_ID, P_TOPIC_ALIAS_MAX, P_TOPIC_ALIAS, P_USER = 0x0B, 0x22, 0x23, 0x26
_U8 = {0x01,0x17,0x19,0x24,0x25,0x28,0x29,0x2A}
_U16 = {0x13,0x21,0x22,0x23}
_U32 = {0x02,0x11,0x18,0x27}
_STR = {0x03,0x08,0x12,0x15,0x1A,0x1C,0x1F}
_BIN = {0x09,0x16}
_VAR = {0x0B}

def vbi(n):
    out = bytearray()
    while True:
        b = n % 128; n //= 128
        if n: b |= 0x80
        out.append(b)
        if not n: return bytes(out)

def rvbi(buf, i):
    mult, val = 1, 0
    for _ in range(4):
        b = buf[i]; i += 1
        val += (b & 0x7F) * mult
        if not b & 0x80: return val, i
        mult *= 128
    raise ValueError("bad varint")

def s(t):
    r = t.encode(); return len(r).to_bytes(2, "big") + r

def rs(buf, i):
    n = int.from_bytes(buf[i:i+2], "big"); i += 2
    return buf[i:i+n].decode("utf-8", "replace"), i + n

def props(items):
    b = bytearray()
    for k, v in items:
        b.append(k)
        if k in _U8: b.append(int(v) & 0xFF)
        elif k in _U16: b += int(v).to_bytes(2, "big")
        elif k in _U32: b += int(v).to_bytes(4, "big")
        elif k in _STR: b += s(str(v))
        elif k in _BIN: b += len(v).to_bytes(2, "big") + v
        elif k in _VAR: b += vbi(int(v))
        elif k == P_USER: b += s(v[0]) + s(v[1])
        else: raise ValueError("prop 0x%02x" % k)
    return vbi(len(b)) + bytes(b)

def rprops(buf, i):
    n, i = rvbi(buf, i); end = i + n; out = {}; users = []
    while i < end:
        k = buf[i]; i += 1
        if k in _U8: v = buf[i]; i += 1
        elif k in _U16: v = int.from_bytes(buf[i:i+2], "big"); i += 2
        elif k in _U32: v = int.from_bytes(buf[i:i+4], "big"); i += 4
        elif k in _STR: v, i = rs(buf, i)
        elif k in _BIN:
            m = int.from_bytes(buf[i:i+2], "big"); i += 2; v = buf[i:i+m]; i += m
        elif k in _VAR: v, i = rvbi(buf, i)
        elif k == P_USER:
            a, i = rs(buf, i); b2, i = rs(buf, i); v = (a, b2); users.append(v)
        else: raise ValueError("unknown prop 0x%02x" % k)
        out[k] = v
    out["users"] = users
    return out, i

class C:
    """One MQTT v5 connection. Every frame is printed so the log IS the evidence."""
    def __init__(self, host, port, tag, verbose=True):
        self.sk = socket.create_connection((host, port), timeout=15)
        self.sk.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
        self.buf = bytearray(); self.tag = tag; self.v = verbose; self.closed = False
        self.stash = []   # packets set aside while waiting for an acknowledgement
        self.log("tcp connect %s:%d" % (host, port))

    def log(self, m):
        if self.v: print("  [%s] %s" % (self.tag, m))

    def send(self, raw, what):
        self.log("--> %-46s %s" % (what, raw[:40].hex()))
        self.sk.sendall(raw)

    def _fill(self, n, deadline):
        while len(self.buf) < n:
            left = deadline - time.monotonic()
            if left <= 0: raise TimeoutError
            self.sk.settimeout(max(0.01, left))
            try: c = self.sk.recv(65536)
            except socket.timeout: raise TimeoutError
            except ConnectionResetError: self.closed = True; raise EOFError
            if not c: self.closed = True; raise EOFError
            self.buf += c

    def read(self, timeout):
        d = time.monotonic() + timeout
        self._fill(1, d); first = self.buf[0]
        n, mult, i = 0, 1, 1
        while True:
            self._fill(i + 1, d); b = self.buf[i]; n += (b & 0x7F) * mult; i += 1
            if not b & 0x80: break
            mult *= 128
        self._fill(i + n, d)
        raw = bytes(self.buf[:i+n]); del self.buf[:i+n]
        p = parse(first, raw[i:]); p["raw"] = raw
        self.log("<-- %-46s %s" % (p["desc"], raw[:40].hex()))
        return p

    def _stashed(self, pred=None):
        """Hand back packets that arrived while an acknowledgement was being awaited.

        A broker may start draining a queue before it answers the SUBSCRIBE that asked for it.
        Discarding those deliveries would make a conformant broker look like it lost them.
        """
        out = []
        while self.stash:
            p = self.stash.pop(0)
            out.append(p)
            if pred is not None and pred(p): break
        return out

    def first(self, pred, seconds):
        """Wait for the first packet matching pred, returning as soon as it arrives.

        Lets a test use a generous deadline without paying for it on a fast broker, which matters
        because delivery latency across implementations here spans two orders of magnitude.
        """
        for p in self._stashed(pred):
            if pred(p): return p
        d = time.monotonic() + seconds
        while True:
            left = d - time.monotonic()
            if left <= 0: return None
            try: p = self.read(left)
            except (TimeoutError, EOFError, ValueError): return None
            if pred(p): return p

    def drain(self, seconds):
        d = time.monotonic() + seconds; out = self._stashed()
        while True:
            left = d - time.monotonic()
            if left <= 0: return out
            try: out.append(self.read(left))
            except (TimeoutError, EOFError, ValueError): return out

    def connect(self, cid, clean=True, se=None, will=None, keepalive=60, extra=None, timeout=20):
        pr = []
        if se is not None: pr.append((P_SESSION_EXPIRY, se))
        pr += (extra or [])
        flags = 0x02 if clean else 0x00
        payload = s(cid)
        if will:
            flags |= 0x04 | ((will.get("qos", 1) & 3) << 3)
            if will.get("retain"): flags |= 0x20
            payload += props(will.get("props", [])) + s(will["topic"])
            payload += len(will["payload"]).to_bytes(2, "big") + will["payload"]
        vh = s("MQTT") + bytes([5, flags]) + keepalive.to_bytes(2, "big") + props(pr)
        body = vh + payload
        self.send(bytes([CONNECT << 4]) + vbi(len(body)) + body,
                  "CONNECT id=%s clean=%d" % (cid, clean))
        p = self.read(timeout)
        assert p["t"] == CONNACK, "expected CONNACK, got %s" % p["desc"]
        return p

    def subscribe(self, filt, opts=1, pid=1, subid=None, timeout=20):
        pr = [(P_SUB_ID, subid)] if subid else []
        body = pid.to_bytes(2, "big") + props(pr) + s(filt) + bytes([opts])
        self.send(bytes([(SUBSCRIBE << 4) | 2]) + vbi(len(body)) + body, "SUBSCRIBE %s" % filt)
        d = time.monotonic() + timeout
        while True:
            p = self.read(d - time.monotonic())
            if p["t"] != SUBACK:
                self.stash.append(p); continue
            if True:
                # SUBACK says the filter was accepted, not that routing already matches it. On a
                # broker with a distributed subscription index it does not yet: BifroMQ answers a
                # publication sent immediately afterwards with PUBACK 0x10, No matching subscribers.
                # Without this pause the control arm of a test can fail for a reason that has
                # nothing to do with the behaviour under test. It is setup, never a measured deadline.
                if all(c < 0x80 for c in p.get("codes", [])):
                    time.sleep(SETTLE)
                return p

    def publish(self, topic, payload, qos=0, pid=None, retain=False, dup=False, pr=None):
        f = (qos & 3) << 1
        if retain: f |= 1
        if dup: f |= 8
        body = s(topic) + (pid.to_bytes(2, "big") if qos else b"") + props(pr or []) + payload
        self.send(bytes([(PUBLISH << 4) | f]) + vbi(len(body)) + body,
                  "PUBLISH %s qos=%d retain=%d dup=%d" % (topic, qos, retain, dup))

    def ack(self, t, pid, reason=0):
        f = 2 if t == PUBREL else 0
        body = pid.to_bytes(2, "big") + (bytes([reason]) + props([]) if reason else b"")
        self.send(bytes([(t << 4) | f]) + vbi(len(body)) + body,
                  "%s pid=%d rc=0x%02x" % (NAMES[t], pid, reason))

    def pub_qos1(self, topic, payload, pid=1, timeout=20, **kw):
        self.publish(topic, payload, qos=1, pid=pid, **kw)
        d = time.monotonic() + timeout
        while True:
            try: p = self.read(d - time.monotonic())
            except (TimeoutError, EOFError): return None
            if p["t"] == PUBACK and p["pid"] == pid: return p
            self.stash.append(p)

    def pub_qos2(self, topic, payload, pid=1, timeout=20, **kw):
        self.publish(topic, payload, qos=2, pid=pid, **kw)
        d = time.monotonic() + timeout
        rec = None
        while rec is None:
            try: p = self.read(d - time.monotonic())
            except (TimeoutError, EOFError): return None
            if p["t"] == PUBREC and p["pid"] == pid: rec = p
        self.ack(PUBREL, pid)
        while True:
            try: p = self.read(d - time.monotonic())
            except (TimeoutError, EOFError): return rec
            if p["t"] == PUBCOMP and p["pid"] == pid: return p

    def disconnect(self, reason=0):
        body = bytes([reason]) + props([])
        try: self.send(bytes([DISCONNECT << 4]) + vbi(len(body)) + body, "DISCONNECT rc=0x%02x" % reason)
        except OSError: pass
        self.close()

    def kill(self):
        """Abnormal close: TCP RST, which section 3.1.2.5 treats as not a normal disconnection."""
        try: self.sk.setsockopt(socket.SOL_SOCKET, socket.SO_LINGER, struct.pack("ii", 1, 0))
        except OSError: pass
        self.log("--- abrupt close (TCP RST)")
        self.close()

    def close(self):
        self.closed = True
        try: self.sk.close()
        except OSError: pass

def parse(first, body):
    t = first >> 4; f = first & 0x0F
    o = {"t": t, "flags": f, "pid": None, "rc": None, "props": {}, "desc": NAMES.get(t, str(t))}
    if t == CONNACK:
        o["sp"] = bool(body[0] & 1); o["rc"] = body[1]
        o["props"], _ = rprops(body, 2)
        o["desc"] = "CONNACK session_present=%d rc=0x%02x" % (o["sp"], o["rc"])
    elif t == PUBLISH:
        o["dup"] = bool(f & 8); o["qos"] = (f & 6) >> 1; o["retain"] = bool(f & 1)
        o["topic"], i = rs(body, 0)
        if o["qos"]: o["pid"] = int.from_bytes(body[i:i+2], "big"); i += 2
        o["props"], i = rprops(body, i); o["payload"] = body[i:]
        o["desc"] = "PUBLISH %r qos=%d dup=%d retain=%d pid=%s payload=%r" % (
            o["topic"], o["qos"], o["dup"], o["retain"], o["pid"], o["payload"][:24])
    elif t in (PUBACK, PUBREC, PUBREL, PUBCOMP):
        o["pid"] = int.from_bytes(body[0:2], "big"); o["rc"] = body[2] if len(body) > 2 else 0
        o["desc"] = "%s pid=%d rc=0x%02x" % (NAMES[t], o["pid"], o["rc"])
    elif t in (SUBACK, UNSUBACK):
        o["pid"] = int.from_bytes(body[0:2], "big")
        o["props"], i = rprops(body, 2); o["codes"] = list(body[i:])
        o["desc"] = "%s pid=%d codes=%s" % (NAMES[t], o["pid"], [hex(c) for c in o["codes"]])
    elif t == DISCONNECT:
        o["rc"] = body[0] if body else 0
        o["desc"] = "DISCONNECT rc=0x%02x" % o["rc"]
    return o

def endpoint(default_port):
    host = sys.argv[1] if len(sys.argv) > 1 else "127.0.0.1"
    port = int(sys.argv[2]) if len(sys.argv) > 2 else default_port
    return host, port

def nid(p): return "%s-%s" % (p, uuid.uuid4().hex[:10])
def ns(): return "repro/%s" % uuid.uuid4().hex[:12]

def verdict(name, failed, detail):
    print("\n" + "=" * 72)
    print("%s: %s" % (name, "REPRODUCED (spec violation observed)" if failed else "not reproduced"))
    print(detail)
    print("=" * 72)
    return 1 if failed else 0

# ---------------------------------------------------------------------------
# TBMQ H-05: a Session whose Session Expiry Interval has elapsed is still resumed
# ---------------------------------------------------------------------------
# Section 3.1.2.11.2: "If the Session Expiry Interval is absent the value 0 is
# used. [...] the Session ends when the Network Connection is closed" -- and a
# granted non-zero interval ends the Session that many seconds after the close.
# Section 3.2.2.1.1: Session Present must be 0 when the Server has no Session
# State for the Client Identifier.
#
# ARM      connect Clean Start 1 with Session Expiry 3, subscribe, DISCONNECT
#          0x00, wait 16s (> 3), reconnect Clean Start 0.
#          Session Present MUST be 0 and the old subscription MUST be gone.
# CONTROL  the same shape reconnecting after 1s, INSIDE the interval, must
#          answer Session Present 1. It proves the broker really creates a
#          Session with this interval, so a 0 in the arm would mean expiry and
#          not "this broker stores nothing".

EXPIRY = 3
WAIT = 16

def main():
    host, port = endpoint(1883)
    topic = ns() + "/a"
    print("target %s:%d  topic %s\n" % (host, port, topic))

    print("-- CONTROL: reconnect INSIDE the granted interval (must be session_present=1)")
    cid = nid("ctl")
    c = C(host, port, "control")
    ack = c.connect(cid, clean=True, se=EXPIRY)
    granted = ack["props"].get(P_SESSION_EXPIRY, "absent (means: as requested)")
    print("   server granted Session Expiry Interval = %s" % granted)
    c.subscribe(topic, opts=1)
    c.disconnect(0)
    time.sleep(1)
    c2 = C(host, port, "control")
    back = c2.connect(cid, clean=False, se=EXPIRY)
    c2.disconnect(0)
    if not back["sp"]:
        return verdict("H-05", False,
                       "Control failed: the broker did not keep the Session even inside its "
                       "interval, so this test cannot say anything about expiry.")
    print("   control OK: session_present=1 inside the interval\n")

    print("-- ARM: wait past the granted interval, then reconnect Clean Start 0")
    xid = nid("arm")
    a = C(host, port, "arm")
    a.connect(xid, clean=True, se=EXPIRY)
    a.subscribe(topic, opts=1)
    a.disconnect(0)
    print("   sleeping %ds, which is %dx the granted interval ..." % (WAIT, WAIT // EXPIRY))
    time.sleep(WAIT)

    a2 = C(host, port, "arm")
    again = a2.connect(xid, clean=False, se=EXPIRY)
    resurrected = False
    if again["sp"]:
        p = C(host, port, "pub")
        p.connect(nid("pub"), clean=True)
        p.pub_qos1(topic, b"after-expiry", pid=1)
        p.disconnect(0)
        for m in a2.drain(6):
            if m["t"] == PUBLISH and m["payload"] == b"after-expiry":
                resurrected = True
                a2.ack(PUBACK, m["pid"])
    a2.disconnect(0)

    return verdict("H-05", bool(again["sp"]),
        "After DISCONNECT 0x00 and %ds (granted Session Expiry Interval was %s), the reconnect\n"
        "answered session_present=%d. %s" % (
            WAIT, granted, int(again["sp"]),
            "The old subscription also still delivered." if resurrected
            else ("The old subscription delivered nothing." if again["sp"] else "Correct.")))

if __name__ == "__main__":
    sys.exit(main())
```

</details>

dmytro-landiak:
## Pull Request description

Fixes #372.

The MQTT 5 Session Expiry Interval was enforced only by the periodic client-session cleanup job (`MQTT_CLIENT_SESSION_EXPIRY_CRON`, hourly by default). A `Clean Start = 0` reconnect between the interval elapsing and the next cleanup run still resumed the session: CONNACK carried `Session Present = 1`, the old subscriptions stayed active and persisted messages were replayed. Per MQTT 5 [3.1.2.11.2] and [3.2.2.1.1] the session must be treated as gone in that case.

Changes:
- `ClientSessionInfo.isExpired(now)`: true when the session is disconnected and `disconnectedAt + sessionExpiryInterval` has elapsed. MQTT 3 `cleanSession=false` semantics (`cleanStart=false`, expiry `0`) never expire by this rule; they remain governed only by the administrative TTL.
- `SessionClusterManagerImpl.updateClientSessionOnConnect`: when the stored session has expired, the connect takes the Clean Start path — subscriptions and persisted messages are cleared, the application-client counter is decremented if applicable, and the CONNACK answers `Session Present = 0`.
- `ClientSessionCleanUpServiceImpl`: reuses the same helper so the cron job and the connect path agree on expiry.

Documentation: none required; behavior now matches the spec and the existing docs.

## General checklist

## Back-End feature checklist


application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/session/SessionClusterManagerImpl.java
```diff
@@ -198,13 +198,26 @@ void updateClientSessionOnConnect(SessionInfo connectingSessionInfo,
...
+    }
+
+    private void processRemoveApplication(SessionInfo connectingSession, ClientSessionInfo currentSession, boolean cleanStart) {
         String clientId = connectingSession.getClientId();
         ClientType clientType = connectingSession.getClientType();
 
         boolean appClientCountDecremented = false;
 
-        if (connectingSession.isCleanStart()) {
+        if (cleanStart) {
```
dmytro-landiak: Could this drift the application-clients counter? `checkApplicationClientsLimit` runs earlier in `processConnectionRequest` and, when the stored session is already a persistent app client and the incoming one is too, it returns `true` *without* incrementing (it assumes the count carries over). With the new expired path, `cleanStart` becomes true here and `decrementAppClientsIfNeeded(currentSession)` fires, so the counter goes down by one while the new persistent app session stays connected and is never counted. Every expired-but-not-yet-cleaned APPLICATION reconnect with Clean Start = 0 would then leak one slot of the quota. The same asymmetry seems to exist today for an explicit Clean Start = 1 reconnect of a persistent app client, but this PR makes it hit the common reconnect case. One option: only decrement when the quota check actually skipped the increment is *not* the case, i.e. pass the expiry decision into `checkApplicationClientsLimit` (treat an expired previous session as `null`) so increment and decrement stay paired.

application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/session/SessionClusterManagerImpl.java
```diff
@@ -198,13 +198,26 @@ void updateClientSessionOnConnect(SessionInfo connectingSessionInfo,
         ));
     }
 
-    private void processRemoveApplication(SessionInfo connectingSession, ClientSessionInfo currentSession) {
+    /**
+     * MQTT 5 [3.1.2.11.2]: the Session ends once the Session Expiry Interval has elapsed after the Network
+     * Connection was closed, so a reconnect must not resume it and must answer Session Present = 0.
+     * The periodic cleanup job may not have removed it yet, hence the check at connect time.
+     */
```
dmytro-landiak: Minor: this wrapper exists only to add a debug log around `currentSession.isExpired(...)`, while the spec reference explaining *why* connect must check lives here rather than at the `cleanStart` computation where the decision is made. Inlining the call and logging inside the existing `if (cleanStart)` block (or a single trace with both flags) would keep the decision and its explanation together.

common/data/src/main/java/org/thingsboard/mqtt/broker/common/data/ClientSessionInfo.java
```diff
@@ -63,6 +64,18 @@ public boolean isDisconnected() {
         return !connected;
     }
 
+    /**
+     * True when this session is disconnected and its MQTT Session Expiry Interval has elapsed.
+     * Sessions with a zero expiry interval and cleanStart=false (MQTTv3 cleanSession=false semantics)
+     * never expire by this rule; they are governed by the administrative TTL only.
+     */
+    public boolean isExpired(long now) {
```
dmytro-landiak: Two things worth a sentence in the javadoc, since this predicate now gates a user-visible decision at connect time rather than an hourly sweep:

1. `disconnectedAt` is stamped by whichever node handled the disconnect and compared against the connecting node's clock. Cross-node clock skew larger than a short Session Expiry Interval (a few seconds) will discard a session that the client believes is still alive. The cron had the same exposure, but connect-time enforcement is far more latency-sensitive.
2. A disconnected session with `disconnectedAt == 0` is unconditionally expired. That is fine for real data (every disconnect path goes through `markSessionDisconnected`), but it is what forced the `disconnectedAt(...)` addition to the `appToDevice` test fixture, and any `ClientSessionInfo` built via `withClientType(...)` or bare builder defaults now silently takes the clean-start path.

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/client/cleanup/ClientSessionCleanUpServiceImpl.java
```diff
@@ -155,27 +147,19 @@ private boolean tryFixGhostSessionState(ClientSessionInfo info) {
...
     }
 
-    boolean isNotCleanSession(ClientSessionInfo sessionInfo) {
-        return sessionInfo.isNotCleanSession();
-    }
-
-    private Long resolveSessionExpiryIntervalMs(ClientSessionInfo session) {
-        if (isNotCleanSession(session)) {
-            return ttl > 0 ? toMillis(ttl) : null;
+    private boolean isExpired(ClientSessionInfo info, long now) {
```
dmytro-landiak: There are now two `isExpired` predicates with different semantics: `ClientSessionInfo.isExpired(now)` deliberately answers false for MQTT3 cleanSession=false sessions, while this private one layers the administrative TTL on top. A reader seeing `info.isExpired(now)` in `SessionClusterManagerImpl` could reasonably assume TTL-expired sessions are also discarded on connect (they are not). Would renaming one make the split explicit (e.g. `isExpiredOrPastTtl` here, or `isSessionExpiryIntervalElapsed` on the domain object)? Alternatively push the TTL branch into the domain object as `isExpired(now, ttlSeconds)` so both callers share one definition and the connect path can opt in later.

application/src/test/java/org/thingsboard/mqtt/broker/actors/client/service/session/SessionClusterManagerImplTest.java
```diff
@@ -266,6 +267,57 @@ public void updateClientSessionOnConnect_noCleanStart_sameType_onlySavesSession(
         verify(applicationRemovedEventService, never()).sendApplicationRemovedEvent(any());
     }
 
+    @Test
+    public void updateClientSessionOnConnect_noCleanStart_previousSessionExpired_clearsSubsAndMsgs_andSavesSession() {
```
dmytro-landiak, line 283: The reported bug is observable end to end (Clean Start = 0 reconnect after the interval elapsed but before the cron ran → Session Present = 1, subs still active), and `SessionExpiryIntegrationTestCase` already has the harness for it, yet the fix is covered only by mocks. Could you add a case there with the cleanup cron effectively disabled (far-future cron via `@TestPropertySource`, or a separate class) that connects with sessionExpiryInterval=1-2s and a subscription, disconnects, waits past the interval, reconnects with cleanStart=false and asserts `token.getSessionPresent() == false` plus an empty `clientSubscriptionCache.getClientSubscriptions(CLIENT_ID)`? The existing `..._whenDisconnectAndReconnectWithoutCleanStart_thenReconnectSuccessful` test runs the cron every second, so it cannot distinguish the connect-time check from the cron and would not have caught #372. Also, none of the three new unit tests assert the CONNACK flag itself (`toSuccessConnectionResponseProto(false)` vs `(true)`), which is the user-visible half of the fix.

application/src/test/java/org/thingsboard/mqtt/broker/actors/client/service/session/SessionClusterManagerImplTest.java
```diff
@@ -241,7 +242,7 @@ public void updateClientSessionOnConnect_cleanStart_clearsSubsAndMsgs_andSavesSe
     public void updateClientSessionOnConnect_appRemoved_appToDevice_clearsMsgs_sendsEvent_andDecrementsIfPersistentAppClient() {
         SessionInfo incomingDevice = deviceSession("appId", false);
         ClientSessionInfo previousApp = ClientSessionInfo.withClientType(ClientType.APPLICATION)
-                .toBuilder().clientId("appId").sessionExpiryInterval(100).build();
+                .toBuilder().clientId("appId").sessionExpiryInterval(100).disconnectedAt(System.currentTimeMillis()).build();
```
dmytro-landiak, line 257: This one-line fixture change is a tell: any `ClientSessionInfo` built without `disconnectedAt` and without `connected(true)` now reads as expired, so fixtures using `ClientSessionInfo.withClientType(...)` or builder defaults silently flip to the clean-start path. A short note on `withClientType` (whose javadoc already says 'for tests') or on `isExpired` would save the next person a confusing test failure.

application/src/test/java/org/thingsboard/mqtt/broker/service/mqtt/client/cleanup/ClientSessionCleanUpServiceImplTest.java
```diff
@@ -126,18 +126,10 @@ public void givenConnectedSession_whenRunCleanup_thenSessionIsNotRemoved() {
...
-        Assert.assertFalse(clientSessionCleanUpService.isNotCleanSession(
-                getClientSessionInfo(false, 100)
-        ));
-        Assert.assertFalse(clientSessionCleanUpService.isNotCleanSession(
-                getClientSessionInfo(true, 0)
-        ));
-        Assert.assertFalse(clientSessionCleanUpService.isNotCleanSession(
-                getClientSessionInfo(true, 100)
-        ));
+        Assert.assertTrue(getClientSessionInfo(false, 0).isNotCleanSession());
```
dmytro-landiak: After removing the `isNotCleanSession` delegate this test no longer exercises anything in `ClientSessionCleanUpServiceImpl`; it is a `ClientSessionInfo` test living in the cleanup service's test class. Move it into the new `ClientSessionInfoTest` or drop it. More importantly, the rewritten TTL branch of the private `isExpired` (`isNotCleanSession` with `ttl > 0` vs `ttl == 0`) has no cleanup-side test: the existing `givenDisconnectedSessionsAndOneOutdated...` case only uses expiry-interval sessions. A `cleanUp()` test with an MQTT3 not-clean session past/inside the TTL, and one with `ttl=0`, seems worth adding since the branch changed in this PR.

application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/session/SessionClusterManagerImpl.java
```diff
@@ -174,8 +174,8 @@ void updateClientSessionOnConnect(SessionInfo connectingSessionInfo,
 
         removeClientLatestTs(clientId);
 
-        boolean cleanStart = connectingSessionInfo.isCleanStart();
-        processRemoveApplication(connectingSessionInfo, currentSession);
+        boolean cleanStart = connectingSessionInfo.isCleanStart() || isPreviousSessionExpired(clientId, currentSession);
```
dmytro-landiak: Behavioural note, not necessarily a change request: an expired previous session now also fires `clearSubscriptionsAndPersist` and `clearPersistedMessages`, which is the right spec outcome, but note `disconnectClientCommandService.disconnectOnSessionConflict(..., connectingSessionInfo.isCleanStart())` at line 485 still receives the raw CONNECT flag rather than the effective `cleanStart`. That path only runs when the old session is still connected (so `isExpired` is false), so it is consistent today; just flagging that the two `cleanStart` notions have diverged if someone later widens `isExpired`.

dmytro-landiak (commented):
## Review summary

Reviewed 6 changed files in [fix(session): discard expired session state on reconnect](https://github.com/thingsboard/tbmq/pull/373). Left 8 comment(s) inline.

The connect-time expiry check is correct for the reported scenario and the cron/connect paths now agree. The main thing to verify is the application-client quota counter on the new expired path (it can be decremented without a matching increment), plus the lack of an end-to-end test that would actually have caught #372.

## Additional findings

- **application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/client/session/SessionSubscriptionServiceImpl.java:76** — `computeSessionEndTs` / `getSessionEndTs` re-implement the same `connected / isNotCleanSession ? ttl : sessionExpiryInterval` branching and `disconnectedAt + seconds→millis` arithmetic that this PR centralises on `ClientSessionInfo`. A single `sessionEndTs(ttlSeconds)` on the domain object (returning -1 for connected / non-expiring) would let the cleanup job, the connect path and the session-details DTO derive from one formula so they cannot drift.

---

application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/session/SessionClusterManagerImpl.java
dmytro-landiak: Confirmed, and the drift also existed for an explicit Clean Start = 1 reconnect of a persistent app client. Fixed in 0b73af04f by pairing the decrement with the quota check instead of touching the check itself: `processRemoveApplication` now releases the slot only when the connecting session is not a persistent APPLICATION client (i.e. only when `checkApplicationClientsLimit` did not carry the slot over). Treating an expired session as `null` in the quota check was rejected because it would increment first and reject the client's own reconnect with QUOTA_EXCEEDED at full quota. Covered by three new `SessionClusterManagerImplTest` cases (expired app→app, Clean Start app→app, app→non-persistent app).

application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/session/SessionClusterManagerImpl.java
dmytro-landiak: Inlined in 0b73af04f: the decision, the spec reference and the debug log now sit together at the `cleanStart` computation; the wrapper is gone.

common/data/src/main/java/org/thingsboard/mqtt/broker/common/data/ClientSessionInfo.java
dmytro-landiak: Both added to the `isExpired` javadoc in 0b73af04f (clock skew across nodes; `disconnectedAt == 0` reads as expired and what fixtures must set). `withClientType` got a matching note.

application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/client/cleanup/ClientSessionCleanUpServiceImpl.java
dmytro-landiak: Went with the third option in 0b73af04f: `ClientSessionInfo.isExpired(now, ttlSeconds)` backed by `getSessionEndTs(ttlSeconds)` is now the single definition. The cleanup job passes the configured TTL, the connect path passes `ClientSessionInfo.NO_TTL` with a comment saying why the administrative TTL does not apply on connect. The private predicate in the cleanup service is removed.

application/src/test/java/org/thingsboard/mqtt/broker/actors/client/service/session/SessionClusterManagerImplTest.java
dmytro-landiak: Added `SessionExpiryOnConnectIntegrationTestCase` in 0b73af04f with the cron set to Feb 29th midnight (Spring cron has no year field, so "far future" is not expressible; this is effectively never within a test run). It connects with expiry 1s + subscription, disconnects, waits 3s, reconnects with cleanStart=false and asserts `token.getSessionPresent() == false` and empty subscriptions, plus a control case inside the interval asserting `true`. Verified RED without the connect-time check. The unit tests now also assert the CONNACK flag: the test producer records responses and the persistence mocks complete their callbacks.

application/src/test/java/org/thingsboard/mqtt/broker/actors/client/service/session/SessionClusterManagerImplTest.java
dmytro-landiak: Documented on both `withClientType` and `isExpired` in 0b73af04f. The fixture change itself stays, since a disconnected session with `disconnectedAt == 0` is, by the domain rule, an expired one.

application/src/test/java/org/thingsboard/mqtt/broker/service/mqtt/client/cleanup/ClientSessionCleanUpServiceImplTest.java
dmytro-landiak: Moved to `ClientSessionInfoTest` in 0b73af04f. Added two cleanup-side tests: MQTT3 not-clean sessions past vs inside the TTL, and `ttl=0` disabling the TTL branch.

application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/session/SessionClusterManagerImpl.java
dmytro-landiak: Agreed, no change. `disconnectOnSessionConflict` only runs when the stored session is connected, where `isExpired` is false by definition (`getSessionEndTs` returns `NO_SESSION_END` for connected sessions), so the raw CONNECT flag is the right input there.

dmytro-landiak:
Re the additional finding on `SessionSubscriptionServiceImpl.computeSessionEndTs`: folded into 0b73af04f. `ClientSessionInfo.getSessionEndTs(ttlSeconds)` is now the single formula (returns `NO_SESSION_END` for connected or non-expiring sessions) and the DTO, the cleanup job and the connect path all derive from it.

application/src/test/java/org/thingsboard/mqtt/broker/service/testing/integration/SessionExpiryOnConnectIntegrationTestCase.java
```diff
@@ -0,0 +1,128 @@
...
+    @Autowired
+    private ClientSubscriptionCache clientSubscriptionCache;
+
+    @Test
+    public void givenExpiredSession_whenReconnectWithoutCleanStart_thenSessionNotPresentAndSubscriptionsCleared() throws Throwable {
+        MqttConnectionOptions options = new MqttConnectionOptions();
+        options.setCleanStart(true);
+        options.setSessionExpiryInterval(1L);
+
+        MqttClient client = new MqttClient(SERVER_URI + mqttPort, CLIENT_ID);
```
dmytro-landiak, line 82: Both tests repeat the same connect → subscribe(MY_TOPIC, QoS 1, no-op listener) → disconnect → close block verbatim, and the second one appends a throwaway `cleanStart=true / expiry=0` connect purely to reset state. Could the setup collapse into a `connectSubscribeAndDisconnect(long sessionExpiryInterval)` helper, with the reset moved into an `@After` so it runs for both? As written the first test also leaves its (expired) session behind and only stays isolated because the second happens to connect with Clean Start = 1.

application/src/test/java/org/thingsboard/mqtt/broker/service/testing/integration/SessionExpiryOnConnectIntegrationTestCase.java
```diff
@@ -0,0 +1,128 @@
...
+        client.subscribe(new MqttSubscription[]{new MqttSubscription(MY_TOPIC, 1)}, new IMqttMessageListener[]{(topic, message) -> {
+        }});
+        client.disconnect();
+        client.close();
+
+        Assert.assertNotNull(clientSessionCache.getClientSession(CLIENT_ID));
+        Assert.assertEquals(1, clientSubscriptionCache.getClientSubscriptions(CLIENT_ID).size());
+
+        // wait past the granted interval; the cleanup cron never fires in this test
+        Thread.sleep(TimeUnit.SECONDS.toMillis(3));
```
dmytro-landiak: A fixed 3 s sleep for a 1 s interval is the kind of slack that either gets trimmed later and goes flaky, or stays and pads the suite. The sibling `SessionExpiryIntegrationTestCase` uses Awaitility throughout; the natural predicate here is `clientSessionCache.getClientSessionInfo(CLIENT_ID).isExpired(System.currentTimeMillis(), ClientSessionInfo.NO_TTL)`, which waits exactly as long as needed and also proves the fixture actually reached the expired state (i.e. `disconnectedAt` was stamped) before the reconnect.

dmytro-landiak (commented):
## Re-review summary

Re-reviewed [fix(session): discard expired session state on reconnect](https://github.com/thingsboard/tbmq/pull/373) — verified 8 finding(s) from [previous review](https://github.com/thingsboard/tbmq/pull/373#pullrequestreview-5264640036).

| Status | Count |
|--------|-------|
| ✅ Resolved | 7 |
| 💬 Acknowledged | 1 |

All previous findings have been addressed. The fix commit (`0b73af04f`) introduces no new correctness issues; the quota pairing in `processRemoveApplication` matches `RateLimitServiceImpl.checkApplicationClientsLimit` in all four combinations (persistent-app → persistent-app carries the slot, everything else increments/decrements symmetrically), and the conflict-disconnect path stamps `disconnectedAt`, so a takeover can never read as expired. Left 7 test-quality comments inline.

### Finding details

- ✅ **application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/session/SessionClusterManagerImpl.java** — app-clients counter drift on expired/Clean Start persistent-app reconnect — *Fixed in code: decrement now guarded by `!connectingSession.isPersistentAppClient()`; 3 new unit tests.*
- ✅ **application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/session/SessionClusterManagerImpl.java** — `isPreviousSessionExpired` wrapper split the decision from its rationale — *Fixed in code: inlined at the `cleanStart` computation.*
- ✅ **common/data/src/main/java/org/thingsboard/mqtt/broker/common/data/ClientSessionInfo.java** — javadoc missing clock-skew and `disconnectedAt == 0` notes — *Fixed in code.*
- ✅ **application/src/main/java/org/thingsboard/mqtt/broker/service/mqtt/client/cleanup/ClientSessionCleanUpServiceImpl.java** — two `isExpired` predicates with different semantics — *Fixed in code: single `ClientSessionInfo.isExpired(now, ttlSeconds)` / `getSessionEndTs(ttl)`, also reused by the session-details DTO.*
- ✅ **application/src/test/java/org/thingsboard/mqtt/broker/actors/client/service/session/SessionClusterManagerImplTest.java** — no end-to-end test with the cron disabled; CONNACK flag not asserted — *Fixed in code: `SessionExpiryOnConnectIntegrationTestCase` added; unit tests assert `sessionPresent`.*
- ✅ **application/src/test/java/org/thingsboard/mqtt/broker/actors/client/service/session/SessionClusterManagerImplTest.java** — `disconnectedAt` fixture tell undocumented — *Fixed in code: note on `withClientType` and `isExpired`.*
- ✅ **application/src/test/java/org/thingsboard/mqtt/broker/service/mqtt/client/cleanup/ClientSessionCleanUpServiceImplTest.java** — orphaned `isNotCleanSession` test; TTL branch untested on the cleanup side — *Fixed in code: moved to `ClientSessionInfoTest`; two cleanup TTL tests added.*
- 💬 **application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/session/SessionClusterManagerImpl.java** — `disconnectOnSessionConflict` still receives the raw CONNECT flag — *Developer: "only runs when the stored session is connected, where `isExpired` is false by definition" — agreed.*

## Additional findings

- **application/src/main/java/org/thingsboard/mqtt/broker/actors/client/service/session/SessionClusterManagerImpl.java:353** (`finishDisconnect` → `markSessionDisconnected`) — pre-existing, not introduced here: an MQTT 5 client that connected with `cleanStart=false, sessionExpiryInterval>0` and then sends DISCONNECT with Session Expiry Interval = 0 is stored as `cleanStart=false, sessionExpiryInterval=0`. That shape is exactly `isNotCleanSession()`, so the unified `getSessionEndTs(NO_TTL)` returns `NO_SESSION_END` and the session never expires on connect (and only via the administrative TTL in the cron), although per [3.1.2.11.2] it should end when the Network Connection closes. A Clean Start = 0 reconnect would still get Session Present = 1. Worth a follow-up issue rather than this PR.

---

application/src/test/java/org/thingsboard/mqtt/broker/service/testing/integration/SessionExpiryOnConnectIntegrationTestCase.java
```diff
@@ -0,0 +1,128 @@
...
+ */
+@Slf4j
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_CLASS)
+@ContextConfiguration(classes = SessionExpiryOnConnectIntegrationTestCase.class, loader = SpringBootContextLoader.class)
+@TestPropertySource(properties = {
+        "mqtt.client-session-expiry.cron=0 0 0 29 2 ?" // effectively never: midnight on Feb 29th
+})
+@DaoSqlTest
+@RunWith(SpringRunner.class)
+public class SessionExpiryOnConnectIntegrationTestCase extends AbstractPubSubIntegrationTest {
```
dmytro-landiak, line 64: The two end-to-end cases cover only the MQTT 5 DEVICE path. Three branches of this fix stop at unit tests although the same harness can drive them: (1) an MQTT v3 `cleanSession=false` reconnect after the sleep must still get Session Present = 1 — that is the `NO_TTL` decision, the one place the unified formula deliberately diverges from the cron; (2) a persistent APPLICATION client whose expired session is discarded on reconnect, asserting the app-client counter does not drift (the new `!connectingSession.isPersistentAppClient()` guard); (3) the PR description says persisted messages are cleared, yet neither test publishes a QoS 1 message while the client is offline and checks it is not replayed after the expired reconnect — the user-visible symptom from #372. Not all three need to land here, but at least (3) seems worth it.

application/src/test/java/org/thingsboard/mqtt/broker/service/mqtt/client/cleanup/ClientSessionCleanUpServiceImplTest.java
```diff
@@ -125,19 +128,35 @@ public void givenConnectedSession_whenRunCleanup_thenSessionIsNotRemoved() {
...
+
+        clientSessionCleanUpService.cleanUp();
+
+        verify(clientSessionEventService).requestClientSessionCleanup(eq(pastTtl), eq(ClientCleanupInfo.GRACEFUL));
+        verify(clientSessionEventService, never()).requestClientSessionCleanup(eq(insideTtl), any());
+    }
+
+    @Test
+    public void givenMqtt3NotCleanSessionAndTtlDisabled_whenRunCleanup_thenSessionIsNotRemoved() {
+        ReflectionTestUtils.setField(clientSessionCleanUpService, "ttl", 0);
```
dmytro-landiak: Reflectively flipping the `@Value` field on a `@MockitoSpyBean` and restoring it in `finally` is fragile — a rename of `ttl` silently turns this into a passing no-op — and the `ttl=0` rule is already pinned by `ClientSessionInfoTest.givenConnectedOrNeverExpiringSession_whenGetSessionEndTs_thenNoSessionEnd`. Since the service now just delegates to `info.isExpired(now, ttl)`, is this test still earning its keep here, or would it read better as a tiny second test class with `ttl=0` in its `@TestPropertySource`?

application/src/test/java/org/thingsboard/mqtt/broker/service/mqtt/client/cleanup/ClientSessionCleanUpServiceImplTest.java
```diff
@@ -51,10 +51,13 @@
 @ContextConfiguration(classes = ClientSessionCleanUpServiceImpl.class)
 @TestPropertySource(properties = {
         "mqtt.client-session-expiry.cron=* * * * * *",
-        "mqtt.client-session-expiry.zone=UTC"
+        "mqtt.client-session-expiry.zone=UTC",
+        "mqtt.client-session-expiry.ttl=60"
```
dmytro-landiak, line 55: `60` is written twice — in the property string and as `TTL_SECONDS` — with nothing tying them together; a mismatch would make `pastTtl` silently land inside the window. Annotation values can't reference the constant, so either read it back via `@Value("${mqtt.client-session-expiry.ttl}")` in the test or leave a one-line comment that the two must stay in sync.

application/src/test/java/org/thingsboard/mqtt/broker/actors/client/service/session/SessionClusterManagerImplTest.java
```diff
@@ -266,6 +279,78 @@ public void updateClientSessionOnConnect_noCleanStart_sameType_onlySavesSession(
...
+    public void updateClientSessionOnConnect_noCleanStart_previousMqtt3NotCleanSession_neverExpires_onlySavesSession() {
+        SessionInfo incoming = deviceSession("c5", false);
+        ClientSessionInfo previous = disconnectedSession("c5", ClientType.DEVICE, false, 0, TimeUnit.DAYS.toMillis(30));
+
+        sessionClusterManager.updateClientSessionOnConnect(incoming, req(), previous);
+
+        verify(clientSessionService).saveClientSession(any(), any());
+        verify(clientSubscriptionService, never()).clearSubscriptionsAndPersist(any(), any());
+        verify(msgPersistenceManager, never()).clearPersistedMessages(any(), any());
+    }
```
dmytro-landiak, line 319: This is the one no-clean-start case in the block that doesn't call `assertConnectResponse(true)`, so Session Present = 1 for the MQTT v3 `cleanSession=false` branch is only implied by the `never()` verifications. Adding it keeps the four sibling tests symmetric and pins the user-visible half of that branch too.

application/src/test/java/org/thingsboard/mqtt/broker/actors/client/service/session/SessionClusterManagerImplTest.java
```diff
@@ -53,12 +55,17 @@
 import org.thingsboard.mqtt.broker.util.ClientSessionInfoFactory;
 
 import java.util.Set;
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
```
dmytro-landiak: Nit: the `java.util.*` imports are now out of order (`Set`, `List`, `CopyOnWriteArrayList`, `UUID`, `TimeUnit`). Same in `ClientSessionCleanUpServiceImplTest`, where `ReflectionTestUtils` landed between the two `org.springframework.test.context.*` imports.

application/src/test/java/org/thingsboard/mqtt/broker/service/testing/integration/SessionExpiryOnConnectIntegrationTestCase.java
dmytro-landiak: Done in 838d8b702: `connectSubscribeAndDisconnect(sessionExpiryInterval)` helper and an `@After` that clears the stored session with a Clean Start = 1 / expiry 0 connect, so every test starts from a clean slate.

application/src/test/java/org/thingsboard/mqtt/broker/service/testing/integration/SessionExpiryOnConnectIntegrationTestCase.java
dmytro-landiak: Replaced with Awaitility on exactly that predicate (`awaitSessionExpired`), which also proves `disconnectedAt` was stamped before the reconnect.

application/src/test/java/org/thingsboard/mqtt/broker/service/testing/integration/SessionExpiryOnConnectIntegrationTestCase.java
dmytro-landiak: Added (3) and (1) in 838d8b702: a QoS 1 message is published while the client is offline and the expired reconnect asserts nothing arrives (Awaitility `during` 2s), plus an MQTT v3 `cleanSession=false` reconnect that must still get Session Present = 1; since that shape has no interval to elapse, the test asserts the broker-side rule directly (`getSessionEndTs(NO_TTL) == NO_SESSION_END`) instead of sleeping. (2) stays at unit level: driving a persistent APPLICATION client end to end needs REST provisioning of the client credentials and the quota config, which is out of proportion for this PR.

application/src/test/java/org/thingsboard/mqtt/broker/service/mqtt/client/cleanup/ClientSessionCleanUpServiceImplTest.java
dmytro-landiak: Dropped. The `ttl=0` rule is pinned by `ClientSessionInfoTest`, and the service is now a pure delegate, so the reflective test added nothing.

application/src/test/java/org/thingsboard/mqtt/broker/service/mqtt/client/cleanup/ClientSessionCleanUpServiceImplTest.java
dmytro-landiak: The test now reads the TTL back via `@Value("${mqtt.client-session-expiry.ttl}")`, so the property string is the only place the value lives.

application/src/test/java/org/thingsboard/mqtt/broker/actors/client/service/session/SessionClusterManagerImplTest.java
dmytro-landiak: Added `assertConnectResponse(true)` to the MQTT v3 not-clean-session case; the four no-clean-start siblings are symmetric now.

application/src/test/java/org/thingsboard/mqtt/broker/actors/client/service/session/SessionClusterManagerImplTest.java
dmytro-landiak: Fixed both files.

application/src/test/java/org/thingsboard/mqtt/broker/service/testing/integration/SessionExpiryOnConnectIntegrationTestCase.java
```diff
@@ -0,0 +1,219 @@
...
+        client.close();
+    }
+
+    @Test
+    public void givenExpiredSession_whenReconnectWithoutCleanStart_thenSessionNotPresentAndNothingReplayed() throws Throwable {
+        connectSubscribeAndDisconnect(1L);
+        awaitSessionExpired();
+
+        // message published while the client is offline; a resumed session would replay it after the reconnect
+        publishQos1("after-expiry");
```
dmytro-landiak, line 96: Is there a timing window here? The broker PUBACKs the QoS 1 publish once it lands in the publish queue, but routing to the DEVICE subscription and writing the persisted message happen asynchronously downstream (two Kafka hops). If the reconnect's `clearPersistedMessages` runs before that write completes but after the routing step matched the still-stored subscription, the message is persisted post-clear and delivered to the reconnected client, and the `received.get() == 0` guard fails spuriously. The publisher `disconnect()`/`close()` plus client construction probably covers it in practice, but a short Awaitility on the persisted-message store for `CLIENT_ID` (or on the subscription count still being 1 and the device queue having 1 entry) before the reconnect would make the fixture deterministic. Worth a look, or have you seen this stable across repeated runs?

application/src/test/java/org/thingsboard/mqtt/broker/service/testing/integration/SessionExpiryOnConnectIntegrationTestCase.java
```diff
@@ -0,0 +1,219 @@
...
+        Assert.assertTrue(token.getSessionPresent());
+        Assert.assertEquals(1, clientSubscriptionCache.getClientSubscriptions(CLIENT_ID).size());
+
+        client.disconnect();
+        client.close();
+    }
+
+    @Test
+    public void givenMqtt3NotCleanSession_whenReconnect_thenSessionPresentAndSubscriptionsKept() throws Throwable {
+        org.eclipse.paho.client.mqttv3.MqttConnectOptions options = new org.eclipse.paho.client.mqttv3.MqttConnectOptions();
```
dmytro-landiak: Nit: five fully-qualified `org.eclipse.paho.client.mqttv3.*` references make this method hard to scan. `MqttConnectOptions` and the v3 `IMqttToken` don't clash with the v5 imports in this file (only `MqttClient` collides), so they can simply be imported; or extract the v3 connect/subscribe/disconnect into a small helper so the FQNs live in one place.

application/src/test/java/org/thingsboard/mqtt/broker/actors/client/service/session/SessionClusterManagerImplTest.java
```diff
@@ -266,6 +279,78 @@ public void updateClientSessionOnConnect_noCleanStart_sameType_onlySavesSession(
...
+    @Test
+    public void updateClientSessionOnConnect_noCleanStart_previousMqtt3NotCleanSession_neverExpires_onlySavesSession() {
+        SessionInfo incoming = deviceSession("c5", false);
+        ClientSessionInfo previous = disconnectedSession("c5", ClientType.DEVICE, false, 0, TimeUnit.DAYS.toMillis(30));
+
+        sessionClusterManager.updateClientSessionOnConnect(incoming, req(), previous);
+
+        verify(clientSessionService).saveClientSession(any(), any());
+        verify(clientSubscriptionService, never()).clearSubscriptionsAndPersist(any(), any());
+        verify(msgPersistenceManager, never()).clearPersistedMessages(any(), any());        assertConnectResponse(true);
```
dmytro-landiak: Nit: `assertConnectResponse(true)` landed on the same line as the `verify(...)` — could you split it onto its own line like the sibling tests above?

dmytro-landiak (commented):
## Re-review summary

Re-reviewed [fix(session): discard expired session state on reconnect](https://github.com/thingsboard/tbmq/pull/373) — verified 7 finding(s) from [previous review](https://github.com/thingsboard/tbmq/pull/373#pullrequestreview-5264804744).

| Status | Count |
|--------|-------|
| ✅ Resolved | 7 |
| 💬 Acknowledged | 0 |

All previous findings have been addressed. The fix commit (`838d8b702`) is test-only and introduces no correctness regressions in the main code paths. Left 3 minor comments inline on the new test code (one possible timing race, two formatting nits).

### Finding details

- ✅ **application/src/test/java/org/thingsboard/mqtt/broker/service/testing/integration/SessionExpiryOnConnectIntegrationTestCase.java** — duplicated connect/subscribe/disconnect block, reset via throwaway connect — *Fixed in code: `connectSubscribeAndDisconnect(long)` helper + `@After clearStoredSession`.*
- ✅ **application/src/test/java/org/thingsboard/mqtt/broker/service/testing/integration/SessionExpiryOnConnectIntegrationTestCase.java** — fixed 3 s `Thread.sleep` — *Fixed in code: `awaitSessionExpired()` polls `isExpired(now, NO_TTL)` with Awaitility.*
- ✅ **application/src/test/java/org/thingsboard/mqtt/broker/service/testing/integration/SessionExpiryOnConnectIntegrationTestCase.java** — end-to-end coverage limited to the MQTT 5 DEVICE happy path — *Fixed in code: offline QoS 1 not-replayed assertion and MQTT v3 `cleanSession=false` case added; the persistent APPLICATION case stays at unit level per the developer's reply (REST provisioning out of proportion) — agreed.*
- ✅ **application/src/test/java/org/thingsboard/mqtt/broker/service/mqtt/client/cleanup/ClientSessionCleanUpServiceImplTest.java** — reflective `ttl=0` flip on a spy bean — *Fixed in code: test dropped; rule pinned in `ClientSessionInfoTest`.*
- ✅ **application/src/test/java/org/thingsboard/mqtt/broker/service/mqtt/client/cleanup/ClientSessionCleanUpServiceImplTest.java** — `60` duplicated between property and constant — *Fixed in code: read back via `@Value("${mqtt.client-session-expiry.ttl}")`.*
- ✅ **application/src/test/java/org/thingsboard/mqtt/broker/actors/client/service/session/SessionClusterManagerImplTest.java** — MQTT v3 not-clean-session case missing `assertConnectResponse(true)` — *Fixed in code.*
- ✅ **application/src/test/java/org/thingsboard/mqtt/broker/actors/client/service/session/SessionClusterManagerImplTest.java** — import order in two test files — *Fixed in code.*

---

application/src/test/java/org/thingsboard/mqtt/broker/service/testing/integration/SessionExpiryOnConnectIntegrationTestCase.java
dmytro-landiak: Real window, even if narrow. Fixed in 5e68d9a4c: after the publish the test awaits `deviceMsgService.findPersistedMessages(CLIENT_ID)` reaching size 1 before reconnecting, so the persisted write is known to have completed and the clear cannot race it. It was stable across the runs I did, but it is deterministic now rather than probabilistic.

application/src/test/java/org/thingsboard/mqtt/broker/service/testing/integration/SessionExpiryOnConnectIntegrationTestCase.java
dmytro-landiak: `MqttConnectOptions` is imported now. The v3 `IMqttToken` does clash (the v5 one is imported in this file), so the v3 client construction and connect moved into a `connectMqtt3` helper returning a small record; the FQNs live only there.

application/src/test/java/org/thingsboard/mqtt/broker/actors/client/service/session/SessionClusterManagerImplTest.java
dmytro-landiak: Split, in 5e68d9a4c.

